Issue #5, opened 2015-12-28
Problem with prompts


ishitatsuyuki:
Prompts other than "steeef" and "eriner" are not working as they are considered.
For example, I got this when I selected "minimal":
```
$(prompt_user)$(prompt_jobs)$(prompt_vimode)$(prompt_status)                                                                                $(prompt_path)$(prompt_git)
```
I don't know why but after this it works like the screenshot:
```
prompt eriner
prompt minimal
```

Eriner:
Just pushed a fix. This has to do with the use of single vs double quotes (literal vs non-literal). Check if it (minimal) works now.

ishitatsuyuki:
Looks nice.

Eriner:
Great. Thanks for the report. Also fixed this for the gitster prompt.

PR #8, opened 2015-12-30
fix prompt substitutions/expansions

caedus75:
This is in relation to #5 

The issue wasn't the single quotes, but the 'prompt_subst' option not being properly set.
This PR fixes the affected themes.

### DIFF
--- a/modules/prompt/themes/gitster.zsh-theme
+++ b/modules/prompt/themes/gitster.zsh-theme
@@ -3,6 +3,9 @@
 # https://github.com/shashankmehta/dotfiles/blob/master/thesetup/zsh/.oh-my-zsh/custom/themes/gitster.zsh-theme
 #
 
+autoload -Uz colors && colors
+prompt_opts=( cr subst percent )
+
 local ret_status="%(?:%{$fg_bold[green]%}➜ :%{$fg_bold[red]%}➜ %s)"
 
 function get_pwd(){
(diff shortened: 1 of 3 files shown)


Eriner:
Ah, you're completely right! This also explains the odd phenomena where switching to the `eriner` prompt would fix these issues (by setting `prompt_opts`). I'm mostly prompt ignorant, so I actually had no idea about this. Thanks for the PR! 

Interesting that switching the quotes also fixes the problem, though. I wonder if that changes when/where the interpretation/swap occurs. I'll debug it later out of my own curiosity. It makes sense, though. It just evaluates the substitutions on each set, whereas setting the prompt_opts causes this substitution to happen to the literal. Interesting.

I will PR momentarily; just want to make some coffee first.

Eriner:
This is super interesting. So according to the [docs](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html), it basically just runs `print -P` on the prompt when those options are set. running the following in a terminal produces visual feedback as to exactly what happens:
```
matt at arch in ~ 
$ print ${PROMPT}

%{%F{135}%}%n at %{%F{166}%}%m in %{%F{118}%}%~ %{%}
$ 

matt at arch in ~ 
$ print -P $_

matt at arch in ~ 
$ 

matt at arch in ~ 
$ 
```

Fantastic. Thanks so much for this PR and teaching me something new!

Eriner:
It's also a shame that there isn't much documentation about `prompt_opts` (at least, to my knowledge). Had to look at the source to even find reference to it.

caedus75:
While searching why `setopt` don't work from within a theme, I found [this thread](http://www.zsh.org/mla/workers/1999/msg03368.html) on the zsh mailing list. They talk about setting prompt options inside themes and the patch that implements `prompt_opts`. It is the best I could find about it.

That's how I figured how to fix the themes.

Eriner:
@caedus75 Yeah, I saw that doing some searches. But seriously, that post is from `Date: Tue, 19 Oct 1999 20:11:22 +0100`

You would think that, at some point, someone would have added that to the documentation someplace. Crazy that it isn't there (or at least not anyplace obvious; I couldn't find it).

Anyway, thanks again for fixing this issue! :+1:

Issue #13, opened 2016-01-05
Themes don't support PROMPT_DIRTRIM


colemickens:
It would be nice if the default themes supported PROMPT_DIRTRIM so that the prompt length can be limited. This is especially useful with the `eriner` theme.

Eriner:
How about just shortening the PWD in general? 

See the `eriner` prompt in this branch:
https://github.com/Eriner/zim/tree/shortpwd

Eriner:
To be clear, this is what I *will* be doing for the `eriner` prompt. The reason I ask is because the issue states: 
> It would be nice if the default theme**s** supported PROMPT_DIRTRIM.

Taking this into account, what I'm really asking is if you would prefer that `${PROMPT_DIRTRIM}` be applied to other (single-line) prompts, or if you would prefer the method I used. Just curious.

colemickens:
I'd prefer to just have the same thing everywhere. What you have in that branch looks nice. Thanks!

Eriner:
Closed by https://github.com/Eriner/zim/commit/be23020b2ef6efd7148c5610b0844c52a826aa4d

PR #14, opened 2016-01-05
add support for PROMPT_DIRTRIM to eriner

colemickens:
Addresses #13 for eriner.


colemickens:
Closing since there already is a branch with a different solution for this.

Issue #12, opened 2016-01-04
prompt module previews strangeness.


beanaroo:
I apologize for the horrible title, I don't really know what's going on, broken prompt module or just a broken theme or two, please rename as appropriate.
I have installed zsh-zim-git via the Arch user repositories. No custom configuration.

Using the `prompt -p` command:

* steeef is not  included in previewed list, depending on the active prompt.
* Preview for liquidprompt is the current active prompt, also seems to depend on active prompt.
* Some prompt previews are broken when previewing after switching to suse theme. I have not confirmed with all the others.
* After running `prompt -p`, active prompt switches to liquidprompt and doesn't change appearance even when changing theme and confirming with `prompt -c`

Eriner:
Yeah, I can verify some of these problems. I'll look into it tomorrow. Likely some prompts are polluting the environment and causing this issue. Thanks for the report!

Eriner:
It looks like will probably be a pain to fix. 

I'll spend some time on it throughout the week, but as this doesn't impair day-to-day functionality and will likely require a hefty time investment to fix, it's not on the top of my to-do list.

If there is anyone who works with prompts frequently (and can quickly determine the issue), I'd be happy to PR.

PR #15, opened 2016-01-06
add setup function to prompt themes

caedus75:
Addresses #12 

Refactor the themes to have a structure similar to zsh default themes.

Include a `prompt_$theme_setup` in all themes and rename functions to prevent them from sharing the same name on different files. This fixes a issue where some information was not being displayed after the themes preview.

To fix `liquidprompt`, I renamed the symlink and added a small wrapper setup to load the prompt and display it when calling `prompt -p`.
It's more a workaround since the theme is an external module.

From my tests, the issues are fixed.
If not, let me know and I'll look into it.

### DIFF
--- a/modules/prompt/functions/ext-liquidprompt
+++ b/modules/prompt/functions/ext-liquidprompt
@@ -1,0 +1,1 @@
+../external-themes/liquidprompt/liquidprompt
--- a/modules/prompt/functions/prompt_liquidprompt_setup
+++ b/modules/prompt/functions/prompt_liquidprompt_setup
@@ -1,1 +1,15 @@
-../external-themes/liquidprompt/liquidprompt
+function prompt_liquidprompt_setup() {
+    autoload -Uz ex-liquidprompt
+
+    ext-liquidprompt
+    prompt_opts=(cr subst percent)
+}
+
+function prompt_liquidprompt_preview() {
+    _lp_set_prompt
+    prompt_preview_theme liquidprompt
+}
+
+prompt_liquidprompt_setup "$@"
+
+# vim:filetype=zsh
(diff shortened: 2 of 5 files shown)


Eriner:
Oh man, you're my prompt-theme hero. Give me a day or two to make sure this fixes everything (going to bed in a few minutes). Thanks for this! :+1:

Eriner:
Also, I may perform an additional commit following this PR to make some additional refactors (mainly POSIX/cosmetic stuff) to the themes. There really shouldn't be any reason for them to be using ksh style `function`s.

Eriner:
I made two commits, one being your original commit, and a second followup commit. I normally would have just amended your commit to include the things I changed, but it was a large amount of changes. Plus, I wanted you to be able to clearly see what I changed. https://github.com/Eriner/zim/commit/9f838982fba465fcfcbfe3fcf345c4edc61d8d8e

Most of these are changes from ksh to POSIX functions, as well as wrapping variables with `${var}` instead of just `$var`; this wasn't anything you changed, just wanted to let you know. Additionally, please don't add vim modelines to the files.

All that said, I really appreciate your work in fixing that issue. I don't think any issues will arise because of these changes; I tested this as much as possible, though I may have missed something.

Thanks again! :+1:

caedus75:
Ooops, sorry about the modeline, wasn't my intention to commit the file with this line, completely forgot to delete it.

With your changes the themes are consistent with the rest of the code. I wasn't sure how to change the style so I just followed the one on each theme. They are looking really good now!

And thank you for zim, it's really great! I'm glad that I can help!
:smile:

Eriner:
Glad you like it! I've laid out a brief style guide [here](https://github.com/Eriner/zim/wiki/Code-Style-Guide), but it's by no means complete. I'll probably flesh it out within the next few days.

PR #32, opened 2016-02-06
Don't call sudo on yaourt

### LINKED ISSUE #31: Don't call certain pacman frontends with sudo
`yaourt` does not require to be called as root, because it asks when root access is required.

Currently, when e.g. `paci` is used, and I wish to manually select packages to install/upgrade, `$VISUAL` must be set. If  `vim`, for example, is entered manually when asked, possibly required directories are not found under `/root` unless manually copied.

This can be fixed by not calling `yaourt` with `sudo`.

Tmplt:
As explained in #31, yaourt does not require to be called as root,
because it requests access by its own.

Closes #31

### DIFF
--- a/modules/pacman/init.zsh
+++ b/modules/pacman/init.zsh
@@ -22,11 +22,15 @@ you can fix this error by editing the 'zpacman_frontend' variable in your .zimrc
   zpacman_frontend='pacman'
 fi
 
+if [[ ${zpacman_frontend} != "yaourt" ]]; then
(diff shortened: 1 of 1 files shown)


Tmplt:
This was the best implementation I could come up with, so that the code's readability could stay the same.

However, if `yaourt` isn't used, `zpacman_frontend` will be `sudo pacman`, which could cause issues if the pacman module would be used elsewhere.

PR #41, opened 2016-03-25
[prompt] fix gitster prompt

cgundogan:
Currently, `gitster's` prompt is not working as intended.
Looking at the code, I'd expect the prompt to show the current path **starting from** the git root (if inside a git managed folder). Otherwise, show the normal output of `short_pwd`.
However, the full (shortened) path is shown instead. The problem is that the previous author mixes `PWD` and `short_pwd`, thus breaking the substitutions.

I added an optional parameter to `short_pwd` and re-use this function to also shorten the output of `git rev-parse --top-level`. This way, the substring matches will succeed.

### DIFF
--- a/modules/prompt/functions/short_pwd
+++ b/modules/prompt/functions/short_pwd
@@ -1,7 +1,9 @@
 # shortens the pwd for use in prompt
 
-local current_dir="${PWD/#${HOME}/~}"
+local current_dir="${1:-${PWD}}"
 local return_dir='~'
+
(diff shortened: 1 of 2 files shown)


Eriner:
@cgundogan Thanks for submitting this PR. I'm a bit busy today, but I'll look at this tomorrow and hopefully merge it. Thanks!

Eriner:
I'm going to go ahead and merge this. I don't use this prompt theme, but your patch changes the behaviour to match what I would expect from a **git**ster prompt. :+1:

PR #48, opened 2016-04-25
fix for https://github.com/Eriner/zim/issues/47

gyakovlev:

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -30,8 +30,15 @@ else
   # colors for ls and completion
   export LSCOLORS='exfxcxdxbxGxDxabagacad'
   export LS_COLORS='di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:'
-
-  alias ls='ls -G'
+ 
+  # stock OpenBSD ls does not support colors at all, but colorls does. 
+  if [[ $OSTYPE == openbsd* ]]; then
+    if (( ${+commands[colorls]} )); then
+      alias ls='colorls -G'
+    fi 
+  else  
+    alias ls='ls -G'
+  fi
 fi
 
 


modules/utility/init.zsh
```diff
@@ -30,8 +30,15 @@ else
   # colors for ls and completion
   export LSCOLORS='exfxcxdxbxGxDxabagacad'
   export LS_COLORS='di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:'
-
-  alias ls='ls -G'
+ 
+  # stock OpenBSD ls does not support colors at all, but colorls does. 
+  if [[ $OSTYPE == openbsd* ]]; then
+    if [[ -x /usr/local/bin/colorls ]]; then
```
Eriner: Please use the `${commands}` array to check for the `colorls` executable.

modules/utility/init.zsh
```diff
@@ -30,8 +30,15 @@ else
   # colors for ls and completion
   export LSCOLORS='exfxcxdxbxGxDxabagacad'
   export LS_COLORS='di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:'
-
-  alias ls='ls -G'
+ 
+  # stock OpenBSD ls does not support colors at all, but colorls does. 
+  if [[ $OSTYPE == openbsd* ]]; then
```
Eriner: Quick question, it seems that FreeBSD has the `ls -G` option. Is OpenBSD the only OS that doesn't provide this?

gyakovlev: I've done a little bit of research on major BSDs.
OSX, FreeBSD and DragonflyBSD have ls -G.
OpenBSD's ls does not have -G.
NetBSD's ls is also colorless, but users have 3! ways of getting color. 

1. install colorls (same trick as for OpenBSD) 
2. install coreutils (which includes ls) 
3. install gnuls

At this point I'm not sure how to deal with all that variety of ls versions. Having conditionals for every os and ls version is ugly.
maybe make a check you are on freebsd/dragonflybsd/osx and append -G, and leave ls alone for the rest of OS and let users deal with colors? This should be fair.
$OSTYPE on dflyBSD matches dragonfly*

Eriner:
@gyakovlev if you fix the first comment I've made in your PR, I'm okay with merging this. Just squash your commits after you make the change please :)

gyakovlev:
@Eriner hey, seems I've fixed colorls detection and was even able to squash the commit. Still learning git, sorry mess I was trying to create =)

modules/utility/init.zsh
```diff
@@ -30,8 +30,15 @@ else
...
+    if (( ${+commands[colorls]} )); then
+      alias ls='colorls -G'
+    fi 
+  else
```
Eriner: This `else` should be outside of this conditional. 
This says: if openbsd and `colorls`, alias `ls`. The `alias ls='ls -G'` will only be hit if openbsd is installed.

gyakovlev: Hey.
Logic is fine. 
`If openbsd and colorls -> alias colorls ; else ls -G`. This gets applied to non-openbsd bsd.
Just tested it on FreeBSD and it works.

Eriner: Oh, I was looking at the commit and misaligned the termination of the `if`s. I didn't realize that the whole thing was in a conditional. This nesting wasn't an issue before due to the previous structure, so I was just looking at that tidbit. You are correct, the logic is fine. Derp. :)

Eriner:
Merged. Thanks for your work!

PR #53, opened 2016-05-05
Small additions to [git] module

hypnoglow:
Just one alias fix and 2 new aliases

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -103,6 +103,7 @@ Aliases
   - `gld` displays the diff log.
   - `glo` displays the one line log.
   - `glg` displays the graph log.
+  - `glG` displays the graph log with authors and dates.
   - `glb` displays the brief commit log.
   - `glc` displays the commit count for each contributor in descending order.
 
@@ -160,6 +161,7 @@ Aliases
   - `gss` stashes the changes of the dirty working directory, including untracked.
   - `gsS` stashes the changes of the dirty working directory interactively.
   - `gsw` stashes the changes of the dirty working directory retaining the index.
+  - `gsu` unapplies (reverts) applied changes.
 
 ### Submodule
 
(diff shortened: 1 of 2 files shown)


Eriner:
the `grc` alias was intentionally avoided to prevent clobbering http://www.unix.com/man-page/debian/1/grc/

hypnoglow:
I would argue on this.

Should ZIM care about distro-specific and user-specific apps?
For example, "ghostscript" package in Archlinux provides `gs` command which runs ghostscript interpreter. But in ZIM `gs` is for `git stash`.

I think ZIM user should care about his naming conflicts, using e.g. `unalias gs` after sourcing ZIM in `.zshrc`.
Another option is to allow aliases longer than 3 symbols, e.g. `grco` for `git rebase --continue`. 

I thinks alias names should be self-explanatory, so I won't try to remember that `grr` is for "git restart rebasing" which is `git rebase --continue`.

Anyway, that's only my thoughts, I'm trying to explore the best solutions, so no offence :)
If you do not agree, I will revert my change and fix `README.md` instead.

Eriner:
It would be nice to determine how many people actually have `grc` ([grcat](https://github.com/pengwynn/grc/blob/master/grcat)) in their path. I agree completely that `grc` would be a better alias.

Eriner:
After giving it some thought, I'm going to merge this. If anyone has `grc` in their PATH, to access the original `grc` command, either:

- `unalias grc` in `.zshrc` after zim init
- use `\grc` to access the "default" (explicitly the unaliased version)

Eriner:
@hypnoglow if you just change the `grc` alias in the Readme, I'll merge this. Thanks!

hypnoglow:
The README is already saying that: "```grc``` continues the rebase after merge conflicts are resolved."

Eriner:
oh, derp. thanks! I thought it didn't and needed to be changed. Makes sense why it's a "fix" now. I'll pull in a moment.

PR #54, opened 2016-05-17
openbsd grep fix - use ggrep or use no color

aplatest:
Minor fix, enabling the SSH module on OpenBSD will break without this fix; OpenBSD grep cannot do colour.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -48,7 +48,13 @@ fi
 
 export GREP_COLOR='37;45'             #BSD
 export GREP_COLORS="mt=${GREP_COLOR}" #GNU
-alias grep='grep --color=auto'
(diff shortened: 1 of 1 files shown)


Eriner:
Can I get someone else who uses OpenBSD to test and confirm this fix? it's a simple one and looks okay at a glance, but I'd like to be sure before I PR. Does this break any greps anywhere else?

gyakovlev:
hey, I discovered the same issue while messing with ls on openbsd.
will test and report back.

Eriner:
@gyakovlev anything to report?

gyakovlev:
@Eriner works perfectly for openbsd without ggrep, and for openbsd with it. Not breaking anything for me.

Eriner:
@gyakovlev thanks for confirming. @aplatest thanks for the PR, merging now!

PR #56, opened 2016-05-21
Added doas-sudo - inserts doas or sudo before current command

aplatest:
Small function that adds sudo (or doas on OpenBSD), before a command. 

Using `!! command` is _not_ an equivalent, as you can use this function _before_ you made the mistake of forgetting sudo.

The short-cut is in line with other zsh short-cuts, like `<esc>h` to open a man-page.

The snippet should follow coding guidelines etc, but let me know if not.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -118,3 +118,20 @@ fi
 mkcd() {
   [[ -n ${1} ]] && mkdir -p ${1} && builtin cd ${1}
 }
+
+# Inserts 'doas ' or 'sudo ' at the beginning of a line.
+function prepend-sudoas {
+  if [[ "$BUFFER" != (doas|su(do|))\ * ]]; then
+    if [[ $OSTYPE == openbsd* ]]; then
+      BUFFER="doas $BUFFER"
+    else
+      BUFFER="sudo $BUFFER"
+    fi
+    (( CURSOR += 5 ))
+  fi
+}
+zle -N prepend-sudoas
+
+# Defined shortcut keys: [Esc] [Esc]
+bindkey "\e!" prepend-sudoas
+


Eriner:
My only issue with this implementation is that I personally spam the `<ESC>` key on occasion. May I suggest a different, more deliberate bind such as `<ESC>!`?

aplatest:
Sure, they're close enough on the keyboard, and if I really hate it I can override it from my `.zshrc`. I'll update my pull-request.

Eriner:
Great, squash those commits, reference this as closing #56 and use the commit format "[input] add doas/sudo insert bind". Thanks!

Eriner:
@aplatest Once you squash the commits (rebase with git, then force push) I'll merge this. Thanks again!

ericbn:
Oh My Zsh and Prezto all became full of specific configurations and customizations and got far away from being a simple and fast set of good defaults for Zsh. I'm just moving from Prezto to Zim and I was happy to see how Zim is a simpler subset of Prezto (I even compared some similar files to find that Zim introduced some few good additions and removed lots of more complex code). Shouldn't Zim be kept slim, avoiding request like this that introduce 17 lines of code just to have a custom variation of what could be done with `sudo !!`... It's cool, of course. But it's also very specific. And is it worth an addition of 17 lines of code?

Eriner:
Since there seems to be some opposition (which has persuaded me), I'm going to hold off on merging this. If there are users who want this functionality, please give the original comment a "thumbs up" so I can get a feel for how many people would find this useful.

ericbn:
Just for reference, this was discussed before in https://github.com/Eriner/zim/issues/16

Tmplt:
Since the changes discussed in #88 seems like the way zim is headed, wouldn't this commit be better fit for an external module (or perhaps a `openbsd` module)?

Eriner:
I think the idea for an `openbsd` module is the best route to take here.

Issue #61, opened 2016-06-22
Parse error related to zdouble_dot_expand with older ZSH


lnicola:
- Zim commit ref: 345233741e441ab00c2ce5540aa00a3646d43abc
- gist of .zimrc: https://gist.github.com/lnicola/50140a843e91eb765e64d882d0093279
- Zsh version: zsh 5.0.2 (x86_64-redhat-linux-gnu)
- gist of .zshrc: https://gist.github.com/lnicola/50140a843e91eb765e64d882d0093279
- $(uname -mosr): Linux 3.10.0-229.1.2.el7.x86_64 x86_64 GNU/Linux

Description
-----------
I get the following error on login:

```
/home/grayshade/.zim/modules/input/init.zsh:68: parse error near `]]'
```

Line 68 is:

```
if [[ ${zdouble_dot_expand} ]]; then
```

Steps to reproduce
------------------
1. Install zim

lnicola:
Perhaps `if [[ ${zdouble_dot_expand} == "true" ]]; then` would be more portable?

PR #62, opened 2016-06-22
Fix parse error with older zsh versions

lnicola:
Fix for issue #61.

### DIFF
--- a/modules/input/init.zsh
+++ b/modules/input/init.zsh
@@ -65,7 +65,7 @@ if [[ -n "${key_info[Insert]}" ]]; then
   bindkey "${key_info[Insert]}" overwrite-mode
 fi
 
-if [[ ${zdouble_dot_expand} ]]; then
+if [[ ${zdouble_dot_expand} == "true" ]]; then
   double-dot-expand() {
     if [[ ${LBUFFER} == *.. ]]; then
       LBUFFER+='/..'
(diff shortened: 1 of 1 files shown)


Eriner:
Yes, this should work. Before, I was just checking if the variable was set at all (uncommented). I'll PR this later this afternoon. Thanks!

lnicola:
But that would be `[[ -n ${var} ]]`, right? The same thing works in newer `zsh` versions, so I think they've extended the syntax for conditionals.

Eriner:
Testing wth just the variable causes a "true" if the variable is set at all (with newer versions of ZSH) and is cleaner than the `-n` syntax, imo. The proposed change works (with the "true" test).

I don't have an issue pulling this, though I would also suggest that you upgrade your version of Zsh if possible. Zsh v5.0.2 was released on Dec 21, 2012. Any compatibility fix for old versions of Zsh that requires a substatial code change (this PR does not) are likely to get a "Update your software, you're running an almost 4 year old shell" from me.

As I said before, I'll merge this later today. Thanks for submitting the PR!

lnicola:
Oh, don't worry, and thanks for including the fix. I didn't realize my zsh version was so old. It's the one in CentOS 7. I only asked because I don't know much about zsh and I never seen that syntax before.

Eriner:
I've merged this PR. Thanks again!

PR #64, opened 2016-07-17
[prompt] add ranger prompt to eriner theme

Tmplt:
I often forget if I'm in a shell under a ranger instance or not, so I prefix an ` r` before the current directory to remind me. While a `ranger` prefix would be expected, it makes the prompt a bit too long, and because it only appears in a shell under ranger, the user would be able to figure out what it means, I believe.

I experimented with including `RANGER_LEVEL` in the prompt also, as it is incremented for every ranger level, but I couldn't figure out how to format it except for `r[0-9-n]`, which I felt was superfluous.

prompt

### DIFF
--- a/modules/prompt/themes/eriner.zsh-theme
+++ b/modules/prompt/themes/eriner.zsh-theme
@@ -72,6 +72,16 @@ prompt_context() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


Eriner:
Merged, thanks!

PR #65, opened 2016-07-24
[steeef] Different prompt for privileged users

aminb:
Use `%(!.#.$)` which will display `$` for normal users and `#` for privileged users


Eriner:
Thanks for the PR, this is a great Quality-of-Life change!

aminb:
Cheers! :beers:

PR #67, opened 2016-07-26
Make sure steeef always updates git information

### LINKED ISSUE #66: steeef theme will not update git status

- Zim commit ref: 5f7a75a969cbfffc6f7e49cffe4be38dc29b9206
- gist of .zimrc: https://gist.github.com/anonymous/d2bc5f9d54ea1c87b527bc7e4443581d
- Zsh version: `zsh 5.1.1 (x86_64-unknown-cygwin)`
- gist of .zshrc: https://gist.github.com/anonymous/c0daff5ad8697298d42876fd9c2c83d1
- $(uname -mosr): `CYGWIN_NT-10.0 2.5.2(0.297/5/3) x86_64 Cygwin`

Description
-----------
The `steeef` theme will not update the git status. You have `cd` out and back in again to update the status. I tested `pure` and `gitster` themes and they both worked as expected. (See screenshot attached)

Steps to reproduce
------------------
Select `steeef`, go into a git repo, change something, profit?

Images or other information
---------------------------
mintty_2016-07-26_15-41-45

This problem exists on my MacBook too, running iTerm and a very recent version of zsh, so it is not a windows/cygwin specific problem.

nshtg:

steeef would not update the git information if we stayed in the same directory. Fixes #66

### DIFF
--- a/modules/prompt/themes/steeef.zsh-theme
+++ b/modules/prompt/themes/steeef.zsh-theme
@@ -38,7 +38,6 @@ prompt_steeef_precmd() {
     zstyle ':vcs_info:*:prompt:*' formats "${FMT_BRANCH} "
 
     vcs_info 'prompt'
-    PR_GIT_UPDATE=
   fi
 
   PROMPT='


nshtg:
I'm not sure what your intention using `PR_GIT_UPDATE` was in the first place, but I don't think only updating the git status on `cd` is a good thing to have.

Eriner:
I was not the creator of this prompt, I merely integrated it into zim. I understand the issues with it only updating the git status when a literal `git` command is called, which is exacerbated by the fact that many people use the git aliases. The ultimate solution here is to update the git repository status like (I think every?) other prompt updates. I'd really like to make them consistent.

nshtg:
Yes agreed. This is just a dirty fix, but in my opinion its enough for a fast hotfix.

Eriner:
Ah, now that I've had time to read the code in context, I approve of this change. From the short diff in GitHub I thought it was just setting an empty variable in an if statement for no reason. I'll PR shortly and close this and #66. Thanks for the fix!

Issue #28, opened 2016-01-28
The cursor disappears when the arrow keys are pressed


RyuzakiKK:
When I write something in zsh and then move the cursor with the arrow keys the cursor disappear until I stop pressing the arrow keys.
Using zsh without zim this issue doesn't appear.

Zsh
![zsh](https://cloud.githubusercontent.com/assets/3227973/12644485/6ef0bb24-c5c3-11e5-8176-15ba784157b2.gif)

Zsh with zim
![zshwithzim](https://cloud.githubusercontent.com/assets/3227973/12644489/74cb1846-c5c3-11e5-9e04-933fff8b182f.gif)

Eriner:
What Terminal emulator are you using? What version of ZSH are you using? What plugins are you using? What is the value of `${TERM}`?

I need information to debug this. Thanks for the report.

GrayShade:
This is my issue from #6. Try disabling the `cursor` highlighter:

```
zhighlighters=(main brackets)
```

Eriner:
That is exactly why I need to know what your environment is, and is also what I suspect. Especially if you're using rxvt, which is a special snowflake. I made note of this in the template .zimrc.

GrayShade:
@Eriner I think there was a confusion in my thread. I had that issue in `gnome-terminal`, while `urxvt` worked fine.

RyuzakiKK:
> What Terminal emulator are you using?
gnome-terminal 3.18.2-1

> What version of ZSH are you using?
zsh 5.2-1

> What plugins are you using?
I simply installed zim (using AUR), no extra plugin chose.

> What is the value of ${TERM}?
xterm-256color

Using the `zhighlighters=(main brackets)` suggested by @GrayShades I solved this problem.

Eriner:
@GrayShade Okay, I thought I remembered experiencing this with urxvt. I will check both to ensure that only gnome-terminal experiences this and then correct the note. Thanks!

Eriner:
Both rvxt and gnome-terminal have this issue. I will update the comment. Thanks again!

Eriner:
Clarifications pushed with https://github.com/Eriner/zim/commit/5cb0a51049eabafb286d1300d5c4efae518c3801.

PR #68, opened 2016-07-27
[template] expand note of highlighter conflicts

nslxndr:
Termite has #28 issue too.


Eriner:
Great, thanks. I'm sure this will save some people frustration down the line! :+1:

PR #69, opened 2016-07-27
[prompt] add new theme: magicmace

Tmplt:
prompt in ~/
prompt in a clean git dir
prompt running under ranger
prompt with a background job
prompt, with last command failing

Dirty git directories are marked with `*`. If the local repo is ahead of behind the remote repo, an arrow is displayed after the branch name. The prompt also tells you if the current branch is diverged. If you are root, the prompt turns red.

I assumed when writing this theme that all I had to do was to drop the files into `modules/prompt/themes/`, but the built-in `prompt` doesn't list it. Do I have to add the theme to a index of some kind?

### DIFF
--- a/modules/prompt/themes/magicmace.zsh-theme
+++ b/modules/prompt/themes/magicmace.zsh-theme
@@ -1,0 +1,100 @@
+#
+# magicmace theme
+# Ideas and code taken from:
+#   xero's zsh prompt <http://code.xero.nu/dotfiles>
+#   eriner's eriner prompt <https://github.com/Eriner/zim/blob/master/modules/prompt/themes/eriner.zsh-theme>
+#
+
+# Global variables
+function {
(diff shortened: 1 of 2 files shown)


modules/prompt/themes/magicmace.zsh-theme
```diff
@@ -0,0 +1,95 @@
...
+
+prompt_git() {
+  local ico_dirty='*'
+  local ico_ahead='🠙'
+  local ico_behind='🠛'
+  local ico_diverged='⥮'
+  local git_info=""
+
+  is_dirty() {
+    test -n "$(git status --porcelain --ignore-submodules)"
```
Eriner: `$(command git status...` is probably what you want. Aliases can cause unexpected issues otherwise.

Eriner:
Thanks for the PR! Merged!

PR #71, opened 2016-08-06
Add lean prompt theme from https://github.com/miekg/lean

fbis251:


Eriner:
I updated the submodule commit and merged it with you. Thanks for the PR!

PR #72, opened 2016-08-15
Add PageUp/Down for navigating through history.

DoctorJellyface:
Incredibly useful when combined history-substring-search and/or a laptop keyboard.

I've tested this and it works fine, and since PageUp/Down isn't binded to anything in Zim I thought it would be a nice addition.

### DIFF
--- a/modules/input/init.zsh
+++ b/modules/input/init.zsh
@@ -61,6 +61,14 @@ if [[ -n "${key_info[End]}" ]]; then
   bindkey "${key_info[End]}" end-of-line
 fi
 
+if [[ -n "${key_info[PageUp]}" ]]; then
+  bindkey "${key_info[PageUp]}" up-line-or-history
+fi
+
+if [[ -n "${key_info[PageDown]}" ]]; then
+  bindkey "${key_info[PageDown]}" down-line-or-history
+fi
+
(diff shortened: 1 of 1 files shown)


Eriner:
My only reservation is that this may be (for the most part) a duplication of the `<UP>` key. An alternative is to make this forcibly override `history-substring-search` and always move to the previously issued command. So, even if you in the middle of a command, ex: `cd foo/bar/long_folder_nam<PAGE_UP>` will forcibly exit you out of that unfinished command and move you back in history to your last command. I hope that makes sense. 

Do you like this idea? Let me know. Either way, yes, the Page Up/Down keys are currently not in use (other than shift+page up/down, which isn't tied to the shell AFAIK).

DoctorJellyface:
> An alternative is to make this forcibly override `history-substring-search` and always move to the previously issued command. So, even if you in the middle of a command, ex: `cd foo/bar/long_folder_nam<PAGE_UP>` will forcibly exit you out of that unfinished command and move you back in history to your last command.

That's exactly how it works now.

Eriner:
Oh geez, I must have read the diff earlier and not remembered I read it. In that case, I will merge this. Thanks!

DoctorJellyface:
Cool, thanks!

PR #76, opened 2016-09-13
Adds colorized-man-pages module

seagoj:
Sets correct variables for less to colorize man pages. Basically implements oh-my-zsh's [colored-man-pages](https://github.com/robbyrussell/oh-my-zsh).

### DIFF
--- a/modules/colorized-man-pages/README.md
+++ b/modules/colorized-man-pages/README.md
@@ -1,0 +1,11 @@
+Colorized Man Pages
+===================
+
+Provides environment variables necessary to add color to `man` pages.
+
+This module will make use of `man`
+
+Functions
+---------
+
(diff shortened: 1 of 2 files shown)


Eriner:
I don't think creating a new module for this functionality is worthwhile, especially since there is no associated README.md file with the module. I would suggest moving this into another module. Additionally, there's no reason to call subshells to `printf`, as that is an expensive operation and it can be done without it. See the Arch wiki about colorized man pages for examples.

If you would like to make those changes I'd be happy to PR this.

seagoj:
I've added a README, moved the man function to the functions directory and implemented it using the method described in Arch wiki.

I don't have an issue combining this with another module, but I'm not sure which one it would make sense to merge it with. Possibly syntax-highlighting?

Eriner:
This functionality was added in commit 7a289c4.

PR #79, opened 2016-09-25
Fix #75: Mac OS X doesn't recognize -o for uname

### LINKED ISSUE #75: zmanage info error on macOS

- Zim commit ref: 071ddaf
- Zsh version: zsh 5.2 (x86_64-apple-darwin15.2.0)
- $(uname -msr): Darwin 15.6.0 x86_64

Description
-----------

On macOS the `uname` command does not accept the option `-o`; which is called from  `~/.zim/tools/zim_info`.

~~~ shell
$ zmanage info
Zim commit ref:   071ddaf
Zsh version:      zsh 5.2 (x86_64-apple-darwin15.2.0)
uname: illegal option -- o
usage: uname [-amnprsv]
System info:
~~~

Steps to reproduce
------------------

~~~ shell
$ uname -msr
Darwin 15.6.0 x86_64
~~~

ericbn:
In `tools/zim_info`, conditionally call `uname` with different parameters depending on the operational system type.

Also fix `uname` calls with `-o` parameter in `trace-zim` and `zim_issue` files. A oneliner was preferred at `trace-zim` to avoid making the code more complex.

### DIFF
--- a/modules/debug/functions/trace-zim
+++ b/modules/debug/functions/trace-zim
@@ -32,10 +32,10 @@ Deleting old logs now..."
 fi
 
 # get some basic system information (kernel and zsh version)
-print "Zsh version: 
+print "Zsh version:
   $(zsh --version)
 Kernel information:
-  $(uname -mosr)
+  $(uname -a)
 fpath info:
   ${fpath}" >! /tmp/ztrace/sysinfo
 
--- a/tools/zim_info
+++ b/tools/zim_info
@@ -6,4 +6,4 @@ cd ${ZDOTDIR:-${HOME}}/.zim
 
 print "Zim commit ref:   $(command git rev-parse --short HEAD)"
 print "Zsh version:      $(command zsh --version)"
-print "System info:      $(command uname -mosr)"
+print "System info:      $(command uname -a)"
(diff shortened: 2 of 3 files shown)


ericbn:
Or maybe `uname -msr` can be used for all OSs? [Solaris](https://docs.oracle.com/cd/E26505_01/html/816-5165/uname-1.html) also apparently does not support `-o`.

Eriner:
I've already made some of these changes in my local branch but have yet to push. I may cherry-pick some of the other changes you've made elsewhere, as I only fixed the issues in `zim_info`. The `uname` command I'm looking for on Mac is `uname -mvsr`, not just `uname -msr`. Regardless, I should have this issue fixed fairly soon. Thanks :+1:

ericbn:
Cool. You could consider using `uname -mvsr` for all OSs to avoid having additional logic for this.

There's a last place I didn't touch, in `modules/debug/functions/trace-zim:88` that may fail in Solaris:

```
  print "your system, ${OSTYPE}, is an unknown system.
Please create an issue at:
  https://github.com/Eriner/zim/issues
Include the following output in your report:
  ${OSTYPE}  -  $(uname -mosr)"
  return 1
```

Eriner:
I've held off on merging this. I think using `uname -a` is the best catch-all that removes all of the additional logic and will always give us what we need. I originally only grabbed the info we needed to debug, but just printing everything reduces code complexity.

If you'd like to update this PR with that change and the change elsewhere, I'll merge it.

ericbn:
Cool. Done. https://github.com/Eriner/zim/pull/84 already got rid of one of the `uname` calls. The 3 remaining ones are being changed in the updated commit.

Eriner:
@ericbn Great, thanks. I'll merge this promptly. :+1:

Eriner:
Merged. Thanks again @ericbn!

PR #80, opened 2016-09-25
Pull latest commit from master for submodules when updating

ericbn:
Instead of using `git submodule update`, that will keep the same commit specified in the containing repository index.

### DIFF
--- a/tools/zim_update
+++ b/tools/zim_update
@@ -8,4 +8,4 @@ cd ${ZDOTDIR:-${HOME}/.zim}
 git remote update -p
 git merge --ff-only @\{u\}
 # and update the submodules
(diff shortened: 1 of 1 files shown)


Eriner:
This looks right. I'll check it later when I get a chance. Thanks!

Eriner:
Essentially reverted by https://github.com/Eriner/zim/issues/127. See ticket there for rational (basically boils down to "ease of use", more advanced users can use `cd ~/.zim && gSu`)

ericbn:
@Eriner, the change proposed in this PR makes the `.zim` repository dirty after submodules are updated, and now I see that this could confuse users. Fair enough.

PR #84, opened 2016-09-27
Refactor debug module to avoid the need of the GNU tools

ericbn:
Use `$PS4` debugging prompt to generate timestamps, instead of GNU `date`. This also keeps timestamps and respective command together, something that `paste` could not guarantee given some commands may have multiple lines.

The prompt timestamp has a precision of 6 decimals. The `printf "%12.9f\n"` instruction with 9 decimals was kept to keep producing "the **SAME** output". This could be replaced to `printf "%12.6f\n"` to make the output smaller. Also, only the first line of commands with multiple lines are saved, to keep the output format backwards-compatible.

### DIFF
--- a/modules/debug/Readme.md
+++ b/modules/debug/Readme.md
@@ -12,16 +12,8 @@ Notes
 -----
 
 The `trace-zim` command will not alter your current dotfiles.
-It will copy your environment to a temporary directory, launch zsh 
-within that environment, and output logs. 
+It will copy your environment to a temporary directory, launch zsh
+within that environment, and output logs.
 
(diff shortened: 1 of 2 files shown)


ericbn:
I'm attaching the output of the refactored code that ran on my Mac OS X for reference: [ztrace.tar.gz](https://github.com/Eriner/zim/files/495936/ztrace.tar.gz)

Eriner:
Great, let me test this out and make one minor additional fix, then I'll merge this. Thanks!

Eriner:
This was a fantastic refactor. Just wanted to say "thanks" one last time! :+1:

PR #89, opened 2016-09-28
Fix #85: Prompt is blank after <Tab><Tab><Ctrl-C>

### LINKED ISSUE #85: Disappearing PS1
Environment Info
----------------
- Zim commit ref: 58aae6a
- Zsh version: zsh 5.2 (x86_64-unknown-linux-gnu)
- Operating System Info: Linux 4.7.5 x86_64 GNU/Linux

Description
-----------

When tabbing through a completion menu, exiting with Ctrl-C while tabbed on the first menu choice results in `${PS1}` disappearing. **This issue is theme-independent**.

Steps to Reproduce
------------------

`cd <tab>`  (press tab until the first menu choice is selected)
<Ctrl-C>

`${PS1}`, AKA "the prompt", will disappear and will not re-draw until a carriage-return (new-line) is entered.

This issue seemingly does not occur if any other menu item is selected, or if no menu item is selected. This issue has existed since zim's inception, but usually I type fast enough and encounter this issue rarely enough that only recently have I discovered the reproducible steps required to trigger this bug.

ericbn:
The issue is fixed by adding `zle redisplay` after `zle expand-and-complete` for the expand-and-complete key binding. An additional `print -n '...'` was also included to show an indicator while the menu is loading, and that is purely cosmetic.

There's still one glitch: the prompt jumps one line up after <kbd>Tab</kbd><kbd>Tab</kbd><kbd>Ctrl-C</kbd>. This was observed in Prezto too.

### DIFF
--- a/modules/input/init.zsh
+++ b/modules/input/init.zsh
@@ -102,6 +102,15 @@ if [[ -n "${key_info[BackTab]}" ]]; then
   bindkey "${key_info[BackTab]}" reverse-menu-complete
 fi
 
+# Redisplay after completing, and avoid blank prompt after <Tab><Tab><Ctrl-C>
+expand-or-complete-with-redisplay() {
+  print -n '...'
+  zle expand-or-complete
+  zle redisplay
+}
+zle -N expand-or-complete-with-redisplay
+bindkey "${key_info[Control]}I" expand-or-complete-with-redisplay
+
 # Put into application mode and validate ${terminfo}
 zle-line-init() {
   if (( ${+terminfo[smkx]} )); then


ericbn:
Or should we call it

    # Redisplay when completing, and avoid blank prompt after <Tab><Tab><Ctrl-C>
    expand-or-complete-with-redisplay() {
      # ...

? Since the redisplay is more important here than the indicator...

Eriner:
@ericbn I like the second function name better, seeing as the actual bug this is fixing is the redispalay.

Eriner:
If you could rename that function, I'll test and hopefully merge this soon. :)

ericbn:
Cool. Commit updated.

PR #99, opened 2016-10-25
[pacman] Add pacI alias to install without syncing

sebastianst:
I often want to install a package without syncing and updating at the same time.
I'd like to see this alias, which I use for a couple of months now locally, merged into zim :)

### DIFF
--- a/modules/pacman/init.zsh
+++ b/modules/pacman/init.zsh
@@ -54,6 +54,9 @@ alias pacb='makepkg -sci'
 # install, sync, and upgrade packages
 alias paci="${zpacman_frontend_priv} -Syu"
 
+# install packages without syncing
+alias pacI="${zpacman_frontend_priv} -S"
+
 # install, sync, and upgrade packages (forcibly refresh package lists)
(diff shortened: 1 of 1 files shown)


sebastianst:
My commit message is wrong and says `pacS` instead of `pacI`. The commit actually adds `pacI` though, as it should... sorry for the typo.

Eriner:
Hm. I'll add this because it doesn't break the [Arch package installation guidelines](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported) since you're not pulling the new repo state. That said, I would still advise the majority of users use `paci` to upgrade package and dependencies against the latest available in their configured repository mirrors.

sebastianst:
Thanks.
Usually, I use `paci`, of course. But it ever so often happens that I install a package and spot one or two optional dependencies that I now want to install too. In this case, I rather use `pacI`. Or 20 min after I did a `pacu` (which I do every day), I want to install that tiny tool, so I just hit `pacI`...

PR #100, opened 2016-10-25
rescope usage to local so it doen't leak out into the shell

Plloi:
Noticed zmanage leaked usage to global scope, so I tagged a quick local on it.


Eriner:
Thanks!

PR #101, opened 2016-10-26
Add less colour, and less or more as default PAGER

ericbn:
In the utility module:
- Set `less` termcap colour variables, and update the module README.md.
- Change the code to enclose all colour logic within a `(( ${terminfo[colors]} >= 8 ))` condition.
- Check if `PAGER` is set before setting the `lm` alias.

In the environment module:
- Set `less` (if available) or `more` as `PAGER`, if `PAGER` is not set.

Keeping the British "colour" spelling. :-)

Closes #76

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -21,7 +21,7 @@ fi
 # Treat single word simple commands without redirection as candidates for resumption of an existing job.
 setopt AUTO_RESUME
 
-# List jobs in the long format by default. 
+# List jobs in the long format by default.
 setopt LONG_LIST_JOBS
 
 # Report the status of background jobs immediately, rather than waiting until just before printing a prompt.
@@ -33,7 +33,16 @@ unsetopt BG_NICE
 # Send the HUP signal to running jobs when the shell exits.
 unsetopt HUP
 
-# Report the status of background and suspended jobs before exiting a shell with job control; 
-# a second attempt to exit the shell will succeed. 
-# NO_CHECK_JOBS is best used only in combination with NO_HUP, else such jobs will be killed automatically
+# Report the status of background and suspended jobs before exiting a shell with job control;
+# a second attempt to exit the shell will succeed.
+# NO_CHECK_JOBS is best used only in combination with NO_HUP, else such jobs will be killed automatically.
 unsetopt CHECK_JOBS
+
+# Set less or more as the default pager.
+if [[ -z ${PAGER} ]]; then
+  if (( ${+commands[less]} )); then
+    export PAGER=less
+  else
+    export PAGER=more
+  fi
+fi
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -3,7 +3,7 @@ Utility
 
 Utility aliases and functions.
 
-Adds colour to `ls` and `grep`.
+Adds colour to `ls`, `grep` and `less`.
 
 Aliases
 -------
@@ -30,7 +30,7 @@ Aliases `get` to ( `aria2c` || `axel` || `wget` || `curl` ).
 
 | alias | command |
 | ----- | ------- |
-| `df` | `df -kh` | 
+| `df` | `df -kh` |
 | `du` | `du -kh` |
 
 ### Condoms
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -3,49 +3,59 @@
 #
 
 #
-# ls Colours
+# Colours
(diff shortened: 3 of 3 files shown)


ericbn:
Sure sure if this is overengineering, but we could also check for `(( ${+commands[less]} ))` before setting it as default...

Plloi:
To me it makes sense to check for the command, less not being the is rare but it does happen.

Eriner:
Checking for the existence of a command is fast, assuming you're using that method (and not something like `$(which less)` which is bad, please don't do that). I would say just do the check, especially since we're changing a variable which may interact with other programs.

In terms of using `tput`, I vote no. Dropping into subshells just to set those sequences seems silly. Subshells are slow -- it's a trap! :)

Eriner:
Though, after thinking about it, I'd rather the `export` not go here, but in `.zshenv`. See: http://zsh.sourceforge.net/Intro/intro_3.html

I'd rather we inform users this kind of thing should be there rather than inserting `exports`. Providing an example `.zshenv` would be a better long-term solution. Give me a day or so to go through and clean up my dotfiles, then I'll push them to GitHub. Hopefully that would be a more informative example.

ericbn:
Sure, `export`s should go to `.zshenv`. On the other hand, the ones in this PR just make sense in an interactive shell.

I see we're also exporting variables in the ssh module (`SSH_AUTH_SOCK`) and the utility module (`LSCOLORS` and `GREP_COLORS`), besides also in some prompts. As for the ones in utility, again they just make sense in an interactive shell, so fine.

And how are we going to provide modularization for the `.zshenv` starting with the new "zimframework" architecture?

ericbn:
Maybe we can move the current changes to the utility module...

Eriner:
Eh, we can do it later. I'm fine with merging this for now. Thanks!

modules/environment/init.zsh
```diff
@@ -33,7 +33,21 @@ unsetopt BG_NICE
...
+# Set less as the default pager.
+[[ -z ${PAGER} && ${+commands[less]} -ne 0 ]] && export PAGER=less
+
+# Termcap for less.
+if [[ ${PAGER} == 'less' ]]; then
+  export LESS_TERMCAP_mb=$'\E[01;31m'     # Begins blinking.
+  export LESS_TERMCAP_md=$'\E[01;31m'     # Begins bold.
+  export LESS_TERMCAP_me=$'\E[0m'         # Ends mode.
+  export LESS_TERMCAP_se=$'\E[0m'         # Ends standout-mode.
+  export LESS_TERMCAP_so=$'\E[00;47;30m'  # Begins standout-mode.
```
Eriner: `$'\E[00;47;30m'` --> `$'\E[01;44;33m'`

Eriner:
Also for a slight structure change, see inspiration from: http://www.zsh.org/mla/users/2007/msg00923.html

While I don't agree with the use of `which`, I think checking for `${terminfo[colors]}` >= 8 makes sense. Anything else I'm forgetting or overlooking? If not, after that change: LGTM.

modules/environment/init.zsh
ericbn: This is what I get with `$'\E[00;47;30m'`:

screen shot 2016-10-28 at 7 27 40 am

And this is what I get with `$'\E[01;44;33m'`:

screen shot 2016-10-28 at 7 28 02 am

ericbn:
Maybe we should also check for `${terminfo[colors]}` >= 8 when setting the `grep` and `ls` colors in the utility module. And then we could also have all these color definitions (for `grep`, `ls`, and for `less`) in the same module...

Another suggestion is we set `more` as the default, if `PAGER` is not previously set and `less` is not present. The `lm` alias in the utility module is assuming `PAGER` is defined and yields a broken alias in case it is not...

ericbn:
And for standout-mode, maybe we should use `$'\E[7m'` [code](http://wiki.bash-hackers.org/scripting/terminalcodes#general_text_attributes) instead of `$'\E[00;47;30m'`.

This is what I get with `$'\E[00;47;30m'` with Solarized Dark and Light:

screen shot 2016-10-28 at 7 27 40 am
screen shot 2016-10-28 at 8 37 49 am

And this is what I get with `$'\E[7m'`:

screen shot 2016-10-28 at 8 35 42 am
screen shot 2016-10-28 at 8 36 13 am

ericbn:
Updated with a new commit containing the changes discusses above.

PR #117, opened 2016-12-24
use setopt NO_CLOBBER instead of unsetopt CLOBBER

ghprince:
zsh has a NO_CLOBBER option, which does exactly the same with unsetopt CLOBBER
https://github.com/zsh-users/zsh/blob/master/NEWS#L1183

use that we will have a clearer explanation.

I also correct the doc.

What NO_CLOBBER does is actually
- prevent `>` to overwrite existing file. use `>|` or `>!` to force.
- prevent `>>` to create non-existing file, use `>>|` or `>>!` to force.


Eriner:
LGTM. Mind squashing your commits and using the commit format on the master branch? If you are busy or unable to, I can take care of it when I merge the branch.

ghprince:
@Eriner how about now?

Eriner:
@ghprince LGTM, thanks!

PR #125, opened 2017-01-17
fix warning when logging in on tty (variable not set)

drm00:
When logging in on the TTY on OpenBSD, the variable $terminfo[colors] is not set, which results in an error while comparing with 8.

Since $terminfo is a read-only variable, i had to take care of the value extraction first before comparing the results, using 0 as the default value if $terminfo[colors] is unset.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -6,7 +6,8 @@
 # Colours
 #
 
-if (( ${terminfo[colors]} >= 8 )); then
+COLORS=${terminfo[colors]}
+if (( ${COLORS:=0} >= 8 )); then
   # ls Colors
   if (( ${+commands[dircolors]} )); then
     # GNU
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -6,8 +6,7 @@
 # Colours
 #
 
-COLORS=${terminfo[colors]}
-if (( ${COLORS:=0} >= 8 )); then
+if [[ ! -z ${terminfo[colors]} ]] && (( ${terminfo[colors]} >= 8 )); then
   # ls Colors
   if (( ${+commands[dircolors]} )); then
     # GNU


modules/utility/init.zsh
```diff
@@ -6,7 +6,8 @@
 # Colours
 #
 
-if (( ${terminfo[colors]} >= 8 )); then
```
Eriner: I believe you can do this inline, `${terminfo[colors]:=0}`

drm00: I tried it like this first, but if $terminfo[colors] is empty, zsh tells me that i can't change a readonly variable, like in this example:

`
$ readonly TEST
$ echo $TEST

$ echo ${TEST:=0}
zsh: read-only variable: TEST
`
Sadly i see no way around using a temporary variable :(

Eriner: In that case, perhaps:

 `if [[ ! -z ${terminfo[colors]} ]] && (( ${terminfo[colors]} >= 8 )); then`

I realize that the `! -z` is a double negative in reality, but I think it makes it more clear *why* we're checking it.

Doing that, it should first ensure that the key-value pair exists, and if it does, then verify that colors >= 8.

drm00: This works, just tried it out. Thanks for finding a better solution ;)

Eriner: @drm00 nope, this was all you, just recommending a small syntax change. If you update your commit, I'll merge this into master. Thanks!

drm00: Should be good now. Thanks!

ericbn:
I've simplified the expression from

    if [[ ! -z ${terminfo[colors]} ]] && (( ${terminfo[colors]} >= 8 )); then

to

    if (( terminfo[colors] >= 8 )); then

at https://github.com/Eriner/zim/commit/6022fe4.

This is preferred and avoids variable substitution according to http://www.bash2zsh.com/zsh_refcard/refcard.pdf

> `var` (does not require `$` in front unless some substitution e.g. `${#var}` is needed, `$` is error if `var` is to be modified)

Without substitution, the undefined variable is defaulted to 0:

```
$ (( ${no_such_array[foo]} >= 8 )) && echo Hi
zsh: bad math expression: operand expected at `>= 8 '
$ (( no_such_array[foo] >= 8 )) && echo Hi
$ (( no_such_array[foo] == 0 )) && echo Hi
Hi
$ typeset -A no_such_array
$ no_such_array=( foo 42 )
$ (( no_such_array[foo] >= 8 )) && echo Hi
Hi
$
```

@Eriner, why did you consider this a bug at https://github.com/Eriner/zim/commit/abbeeb9? The changes there bring this issue back.

ericbn:
I get the same behaviour as above in
* zsh 5.4.2 (x86_64-apple-darwin15.6.0) *macOS*
* zsh 5.0.8 (x86_64-apple-darwin15.0) *macOS*
* zsh 5.4.2 (arm-unknown-linux-androideabi) *Termux*
* zsh 5.1.1 (x86_64-ubuntu-linux-gnu) *Ubuntu on Windows Subsystem for Linux*

PR #128, opened 2017-01-18
Add insertmode indicator

Valinora:
This adds the insertmode indicator to all except the minimal theme because it already has it.


Eriner:
@Valinora give me a ping when you can clean up the commit history a bit and would like me/others to take a look. Thanks for the PR, looking forward to being able to merge this at some point! :+1:

Tmplt:
This doesn't seem to be optional. Perhaps a check on an env variable (`ZIM_PROMPT_INSERTMODE=` `1` or `0`) could be added to toggle the `RPROMPT`?

Valinora:
That's a consideration. I'm likely to re-implement what I did [here](https://github.com/miekg/lean/pull/16).

PR #130, opened 2017-01-21
Add insertmode indicator for default prompts

Valinora:
I apologize my terribleness with git and the need for me to create a new pull request.

This creates a RPROMPT variable that is enabled by the ZIM_PROMPT_INSERTMODE variable's existence. It then binds a zle widget to redraw the prompt with an indicator depicting command mode.


ericbn:
Instead of having repeated code across the prompt themes, I believe we should implement this in a module. Prezto, for example, has implemented theming features in its editor module. See https://github.com/zsh-users/prezto/tree/master/modules/editor#theming

Eriner:
Because we won't be merging this as-is, I'm going to close this PR. We can use this as a reference moving forward.

Issue #111, opened 2016-12-09
fasd completion not working


justbur:
- Zim commit ref: b55ed8ad27e481891d6f0811eed67d92316c5377
- gist of .zimrc: https://gist.github.com/justbur/d90f83a17186a186f228bc60dc5820fa
- Zsh version: 5.2
- gist of .zshrc: (just sources .zim/init.zsh)
- $(uname -mosr): Linux 4.8.11-300.fc25.x86_64 x86_64 GNU/Linux

Description
-----------

It appears that the fasd module sets up completion for the fasd commands. It doesn't appear to be working for me. 

Steps to reproduce
------------------

 1. `z [beginning of some dir]<tab><tab>` has no effect regardless of what I put in for the name of the directory.

Images or other information
---------------------------

spencersteers:
Would love to take a stab at this - anyone familiar enough to know where to start?

PR #132, opened 2017-01-22
Fix fasd completion

deathbeam:
- Add fasd as gitmodule
- Regenerate fasd configuration script from current fasd version
- Fix initialization script to have all fasd-related functions inside it
- Add another alias `v` for fasd that will access recently used files
  from .viminfo

Fixes: #111

Signed-off-by: Tomas Slusny <slusnucky@gmail.com>


Eriner:
Not going to merge this. Upstream fasd is a huge, bloated shell script that isn't zsh specific; we only need the zsh things. I appreciate your contribution, however I would much rather see this fix happen in the current codebase.

deathbeam:
I can then alter this patch to not use upstream fasd, but included one in repo.

Eriner:
@deathbeam I've merged your patch, thanks for the fix. And yes, my rational is as above. The `sh` fasd binary is just full of unnecessary stuffs for zsh, and especially since the repo hasn't seen much action, I thought it made more sense to strip out the unnecessary code so it would be easier to maintain.

PR #133, opened 2017-01-22
Fix fasd completion (no submodule)

deathbeam:
As @Eriner said in  #132, this is fix without adding `fasd` as submodule. 

- Fix word completion for fasd
- Fix initialization script to have all fasd-related functions inside it
  (functions was not called properly when they was in separate files)
- Add another alias `v` for fasd that will access recently used files
  from .viminfo

Fixes: #111


Eriner:
Fantastic, thanks @deathbeam, this is exactly what I was looking for!! Very much appreciated!

PR #142, opened 2017-03-01
Rewrite eriner theme using the Zim git-info module

ericbn:
Based on my rewrite of the Agnoster theme, uses simpler segment drawing functions and only relies on local variables.

Uses git-info module in non-verbose mode, so a repo with only untracked files will not count as dirty (but dirty state is computed faster). Also took the freedom of adding an extra space character after the 'r' for the Ranger segment, and after the dirty indicator for the Git segment. Everything else looks and behaves exactly the same as the original prompt.

Change yields 37 less lines of code.

### DIFF
--- a/modules/prompt/themes/eriner.zsh-theme
+++ b/modules/prompt/themes/eriner.zsh-theme
@@ -1,14 +1,12 @@
+# vim:ts=2 sw=2 sts=2 ft=zsh
+#
 # Eriner's Theme - fork of agnoster
 # A Powerline-inspired theme for ZSH
 #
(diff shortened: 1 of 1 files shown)


ericbn:
Before and after:

screen shot 2017-03-01 at 6 03 10 pm

ericbn:
Something inspired in the magicmace prompt is having the Ranger indicator in the same segment as the error and background job indicators, instead of in separate segment of its own. This is what it would look like (before and after):

screen shot 2017-03-01 at 5 53 58 pm

PR #148, opened 2017-03-09
Inline is-true function body, as other functions

ericbn:
defined in Zim. Instead of creating an additional `is-true() { }` function scope.


Eriner:
Good catch. :+1:

ericbn:
Cool. Change done!

PR #152, opened 2017-03-15
Rewrite minimal theme using git-info

ericbn:
- Add `diverged` context to git-info that, when defined, will be set if branch is both ahead and behind of remote. If not defined, the `ahead` and `behind` contexts will still be set, as how they worked previously.
- Remove undocumented git-info configuration `prompt.showinfo` for enabling or disabling it globally or per repository. It can be globally disabled by not loading the `git-info` module at all, and prompts currently will not break (and in the future should still not break) if the module is not loaded. This removes one git call that is used to check for the `prompt.showinfo` configuration value.
- Add `coalesce` function to git-info that will be used by the minimal theme, and could be used for future themes too. Prezto had this, but we diminished its necessity by simplifying how git-info works.
- Finally, rewrite minimal theme using git-info. As previous migrations of themes to git-info, this does not use its verbose mode, so a repo with just untracked files is not considered dirty. It uses the newly-introduced `diverged` context and `coalesce` function from git-info.

### DIFF
--- a/modules/git-info/README.md
+++ b/modules/git-info/README.md
@@ -48,16 +48,22 @@ a style is:
 | action    |   %s   | Special action name (see Special Action Contexts below)
 | ahead     |   %A   | Commits ahead of remote count
 | behind    |   %B   | Commits behind of remote count
+| diverged  |   %V   | Diverged commits (both ahead and behind are yield when it's not defined)
 | branch    |   %b   | Branch name
 | commit    |   %c   | Commit short hash (when in 'detached HEAD' state)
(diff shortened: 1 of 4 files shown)


ishitatsuyuki:
By the way, your migration to git-info doesn't enable it by default, so it was surprising to me which updated and found the git prompts disappeared. I recommend you to enable it in the default zimrc too, as there's a possibility that new user will never discover the feature.

ericbn:
@ishitatsuyuki, good point. Created https://github.com/Eriner/zim/pull/153 with this change.

PR #155, opened 2017-03-26
Prevent directory change to .zim after running zmanage update

janclarin:
Should be able to run `zmanage update` without being redirected into the .zim folder after it is finished running

**Change**
Return to the previous directory after updating from upstream and updating submodules

### DIFF
--- a/tools/zim_update
+++ b/tools/zim_update
@@ -11,4 +11,4 @@ git merge --ff-only @\{u\}
 git submodule update --init --recursive
 
 # return to previous directory after update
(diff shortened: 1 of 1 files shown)


ericbn:
I'm not "being redirected into the .zim folder after [`zmanage update`] is finished running".

janclarin:
It must've happened by chance. Then again, it was pretty late last night when I encountered this, I'm not able to reproduce it anymore. Sorry, I will close this.

PR #156, opened 2017-03-29
[git] git-commit-lost: fix typo

sebastianst:
command is written with two m, not three.


ericbn:
Good catch! LGTM
(speaking of typos, it's "git-commit-lost", not "git-commit-list" 😄)

sebastianst:
Indeed, it should read `git-commit-lost` ;)

PR #157, opened 2017-03-30
Rewrite steeef theme using git-info

ericbn:
- Change git-info verbose mode implementation so it behaves alike the non-verbose mode. For example, in the "merge" special action context, files are being reported as both indexed and unindexed by `git diff-index` and `git diff-files` commands in non-verbose mode. That was not the case with the regular expressions used in the verbose mode.

- Rewrite steeef theme using git-info. This uses the verbose mode to be able to display indexed, unindexed and untracked files indicators. This is the last theme left to be migrated to git-info, and the first one to use its verbose mode. Originally, the theme also supported svn via vcs_info. The svn support is lost in this rewrite. Maybe if a new svn-info module is introduced to Zim...

### DIFF
--- a/modules/git-info/README.md
+++ b/modules/git-info/README.md
@@ -100,5 +100,5 @@ Second, format how the above attributes are displayed in prompts:
       'rprompt' '[%R]'
 
 Last, add `$git_info[prompt]` to `$PROMPT` and `$git_info[rprompt]` to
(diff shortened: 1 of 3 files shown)


Eriner:
LGTM

Eriner:
And for reference, I'm all if favor of allowing PRs for SVN support if someone wants it, but I'm more than okay removing it until it's needed/a PR is created -- I don't know anyone who is still using SVN (that isn't under a rock).

PR #160, opened 2017-04-01
fix url quote magic

sodiumjoe:
`url-quote-magic` automatically escapes characters that fit a URI scheme
as you type. From what I can tell from the commit history, this was
mistakenly placed in `else` branch of the nested conditional instead of
at the end of the outermost conditional.

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -13,9 +13,9 @@ if [[ ${ZSH_VERSION} != 5.1.1 ]]; then
       autoload -Uz bracketed-paste-magic
       zle -N bracketed-paste bracketed-paste-magic
     fi
-    autoload -Uz url-quote-magic
-    zle -N self-insert url-quote-magic
   fi
+  autoload -Uz url-quote-magic
+  zle -N self-insert url-quote-magic
 fi
 
 # Treat single word simple commands without redirection as candidates for resumption of an existing job.


ericbn:
`bracketed-paste-magic` is known to be [buggy in 5.1.1](https://github.com/robbyrussell/oh-my-zsh/pull/4473), and `url-quote-magic` seems to [depend on it](https://github.com/zsh-users/zsh/blob/a9df6aaa702abf761b155cd842a7f6917be44139/Functions/Zle/url-quote-magic#L11-L14). These are the reasons behind the current logic there.

Eriner:
@ericbn is exactly right.

sodiumjoe:
Thanks for those links, that helps clear things up a bit, but I think the current logic is still incorrect. [The current logic in oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/misc.zsh) is:

```
if [[ ${ZSH_VERSION} != 5.1.1 ]]; then
  if is-at-least 5.2; then
    autoload -Uz bracketed-paste-url-magic
    zle -N bracketed-paste bracketed-paste-url-magic
  else
    if is-at-least 5.1; then
      autoload -Uz bracketed-paste-magic
      zle -N bracketed-paste bracketed-paste-magic
    fi
  fi
  autoload -Uz url-quote-magic
  zle -N self-insert url-quote-magic
fi
```

which is identical to my diff.

sodiumjoe:
the logic as i understand it is:

1. don't use `url-quote-magic` or `bracketed-paste-url-magic` or `bracketed-paste-magic` in `5.1.1`
2. for `>=5.2` use `url-quote-magic` and `bracketed-paste-url-magic`
3. for `>=5.1` and `<5.2` use `url-quote-magic` and `bracketed-paste-magic`

I'm running `5.3.1` and cannot reproduce the bug with `url-quote-magic` and -bracketed-paste-url-magic` enabled, so I think that's correct.

ericbn:
@sodiumjoe, the current login in Oh-My-Zsh is not identical to what you're prosing. **BUT** I think you're right about the logic above, and your changes make sense.

Current code only uses `url-quote-magic` when `<5.2`.

@Eriner, the changes LGTM.

sodiumjoe:
you're right, it's not identical, my mistake

ericbn:
Current logic only uses `url-quote-magic` when `<5.2`, and looks like this needs to be fixed. The proposed changes look better, but still does it make sense to use `url-quote-magic` without any of `bracketed-paste-url-magic` or `bracketed-paste-magic` when `<5.1`?

sodiumjoe:
I was trying to follow the logic in https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/misc.zsh as closely as possible, which does appear to

> use url-quote-magic without any of bracketed-paste-url-magic or bracketed-paste-magic when <5.1

```
if [[ $ZSH_VERSION != 5.1.1 ]]; then
  for d in $fpath; do
  	if [[ -e "$d/url-quote-magic" ]]; then
  		if is-at-least 5.1; then
  			autoload -Uz bracketed-paste-magic
  			zle -N bracketed-paste bracketed-paste-magic
  		fi
  		autoload -Uz url-quote-magic
  		zle -N self-insert url-quote-magic
      break
  	fi
  done
fi
```

PR #162, opened 2017-04-03
Remove git-info dependency on is-true function

### LINKED ISSUE #161: git-info depends on git module
I'd like to use the `git-info` module without the git aliases provided by the `git` module, but if I try disabling `git` zim tells me `git-info:9: command not found: is-true`.

I found that `is-true` is defined inside the `git` module, so it is a dependency to `git-info`.

Shouldn't this definition be higher up the chain so these modules are decoupled?

ericbn:
from `git` module. Actually, the function is a little bit of over-engineering and it's currently only being used for checking the result of `git rev-parse --is-inside-work-tree`, even in the `git` module code.

Checking for the return code of `git rev-parse --is-inside-work-tree` is enough. Also refactored other portions of the code where return codes are being checked.

Fixes #161

### DIFF
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -96,7 +96,7 @@ git-info() {
   typeset -gA git_info
 
   # Return if not inside a Git repository work tree.
-  if ! is-true $(git rev-parse --is-inside-work-tree 2>/dev/null); then
+  if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
     return 1
   fi
 
@@ -230,9 +230,7 @@ git-info() {
   zstyle -s ':zim:git-info:clean' format 'clean_format'
 
   local dirty
-  local indexed
   local indexed_formatted
-  local unindexed
   local unindexed_formatted
   local untracked_formatted
   if ! zstyle -t ':zim:git-info' verbose; then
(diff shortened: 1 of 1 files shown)


ericbn:
Should we do the same change
```
-  if ! is-true $(git rev-parse --is-inside-work-tree 2>/dev/null); then
+  if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
```

in the `git` module, and get rid of the `is-true` function there?

The only places where the function is being used are in that module, and in calls similar to the above.

Eriner:
> Should we do the same change ... in the git module, and get rid of the is-true function there?

Since that would remove the dependency of that function, I'm all for it. In that case, since only prompts (for compatability) would rely on `is-true` AFAIK, I think we may be able to move that function to the `git-info` module as well, correct?

ericbn:
> Since that would remove the dependency of that function, I'm all for it.

Cool. I'll create a separate PR for the `git` module changes, removing the dependencies of `is-true` there.

> I think we may be able to move that function to the `git-info` module as well, correct?

I'm more for removing the `is-true` function completely. If someone raises an issue because one misses it, we can check what's the best place to put it.

Eriner:
LGTM

PR #163, opened 2017-04-11
Remove git dependency on is-true function

ericbn:
- The function is currently only being used for checking the result of `git rev-parse --is-inside-work-tree`. Checking for the return code of the command is enough.
- Remove is-true function that was only being used in the `git` and `git-info` modules to check the output of `git rev-parse --is-inside-work-tree`. The code depending on this function was already changed.


ericbn:
https://github.com/Eriner/zim/pull/162 must be merged before this one, as this removes the `is-true` function.

PR #164, opened 2017-04-12
[git-info] Remove dependency on git-dir function

### LINKED ISSUE #161: git-info depends on git module

ericbn:
from the git module. Also now making all git calls with `command`.

Fixes #161 (again)

### DIFF
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -6,7 +6,7 @@
 # Gets the Git special action (am, bisect, cherry, merge, rebase).
 # Borrowed from vcs_info and edited.
 _git-action() {
-  local git_dir=$(git-dir)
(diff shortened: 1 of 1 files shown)


Eriner:
> Also now making all git calls with `command`.

I had wondered if we had other instances of this laying around, I believe I even mentioned it in a thread somewhere where we first encountered the conflict. Ha! Glad to see it resolved! :+1:

Eriner:
LGTM

Issue #158, opened 2017-03-30
Git-info vulnerable to arbitrary command execution


ericbn:
As reported in https://github.com/sorin-ionescu/prezto/issues/1267:

> `git-info` seems to not do proper escaping when retrieving the branch name. For example, setting the branch to `./an_executable` will run the local executable. An example repo can be found [here](https://github.com/njhartwell/pw3nage).

Actually, when branch is named like `$(./an_executable)`. Need to check how our version of git-info and our prompts are reacting to this.

Eriner:
The "Injection exploit" title from the issue you linked is a bit sensationalized. I guess _technically_, but we're dealing with a shell environment here; it's all pure "command execution" anyway.

PR #168, opened 2017-04-17
[prompt] Don't expand git_info in eriner theme

ericbn:
to avoid the git-info vulnerability to execute an arbitrary command, as reported in #158.

Also, change range segment in eriner theme to show as in magicmace. As suggested in https://github.com/Eriner/zim/pull/142#issuecomment-283806756 (and "thumb-upped" by @Tmplt). The Ranger segment was introduced in eriner theme by @Tmplt in #64. He also added it to the magicmace theme.


Eriner:
LGTM

PR #170, opened 2017-04-24
Add comment to ssh module for GNOME users

MajorFeels:
I felt like I should add this because I ran into this issue. I was thinking this might help other users in my situation.

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -1,6 +1,12 @@
 #
 # sets up ssh-agent
 #
+# The GNOME Keyring daemon will cause issues with ssh-agent
+# Please make sure to disable the GNOME Keyring SSH intergration if you wish to use ssh-agent
+# This can be done with the following commands: 
+#   mkdir ~/.config/autostart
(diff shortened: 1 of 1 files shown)


modules/ssh/init.zsh
```diff
@@ -1,6 +1,12 @@
 #
 # sets up ssh-agent
 #
+# The GNOME Keyring daemon will cause issues with ssh-agent
+# Please make sure to disable the GNOME Keyring SSH intergration if you wish to use ssh-agent
+# This can be done with the following commands: 
+#   mkdir ~/.config/autostart
+#   cp /etc/xdg/autostart/gnome-keyring-ssh.desktop ~/.config/autostart/ &&
+#   printf '%s\n' 'Hidden=true' >> ~/.config/autostart/gnome-keyring-ssh.desktop
```
Eriner: As we're a toolkit for zsh, I suggest:
```
print 'Hidden=true' >> ~/.config/autostart/gnome-keyring-ssh.desktop
```

Issue #176, opened 2017-05-09
git-info ahead/behind is broken


Tuurlijk:
```
Zim commit ref:   a7e060a
Zsh version:      zsh 5.0.8 (x86_64-apple-darwin15.0)
System info:      Darwin Food.local 15.6.0 Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64 x86_64
```

- gist of .zimrc:
https://github.com/Tuurlijk/vim-config/blob/master/.zimrc

- gist of .zshrc:
https://github.com/Tuurlijk/vim-config/blob/master/.zshrc

Description
-----------
When building a promt that accepts colours (like the fire prompt), I noticed that ahead/behind info in the git-info module returned incorrect results

Steps to reproduce
------------------
Enable a prompt with git-info e.g. eriner with the config:

```
	zstyle ':zim:git-info:ahead' format '↑%A'
	zstyle ':zim:git-info:behind' format '↓%B'
	zstyle ':zim:git-info:keys' format \
		'prompt' '%b%c%s%A%B %C%D'
```

Images or other information
---------------------------
On a branch that is behind you will see:
![schermafbeelding 2017-05-09 om 23 18 26](https://cloud.githubusercontent.com/assets/790979/25873216/eafbbc04-350d-11e7-9307-7f8c15ff1b87.png)
Where you should see:
![schermafbeelding 2017-05-09 om 23 18 48](https://cloud.githubusercontent.com/assets/790979/25873221/f127e8be-350d-11e7-9e96-85a5fef584e0.png)

ericbn:
Can you reproduce this with `magicmace` or `minimal` prompt themes?

ericbn:
I'm unable to reproduce this using any of the existing Zim prompts, not by making the changes in eriner as per "Steps to reproduce above". I always get the `↓n` to show in the prompt...

Tuurlijk:
![magicmace](https://cloud.githubusercontent.com/assets/790979/26032701/c8c98de8-3899-11e7-836a-f0454115b4eb.png)
![pure](https://cloud.githubusercontent.com/assets/790979/26032702/cc9c8ea2-3899-11e7-96e4-afeb5c5279f6.png)

ericbn:
I'm unable to reproduce.

[...]

PR #177, opened 2017-05-09
Update git-info/init.zsh

Tuurlijk:
Enable proper ahead / behind behavior

Fixes: #176

### DIFF
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -178,7 +178,8 @@ git-info() {
       local ahead_and_behind_cmd='command git rev-list --count --left-right HEAD...@{upstream}'
 
       # Get ahead and behind counts.
-      local ahead_and_behind=$(${(z)ahead_and_behind_cmd} 2>/dev/null)
+      local ahead_and_behind      
+      ahead_and_behind=$(${(z)ahead_and_behind_cmd} 2>/dev/null)
       local ahead=${ahead_and_behind[(w)1]}
       local behind=${ahead_and_behind[(w)2]}
 
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -179,7 +179,7 @@ git-info() {
 
       # Get ahead and behind counts.
       local ahead_and_behind      
-      ahead_and_behind=$(${(z)ahead_and_behind_cmd} 2>/dev/null)
+      ahead_and_behind=${(z)$(${ahead_and_behind_cmd} 2>/dev/null)}
       local ahead=${ahead_and_behind[(w)1]}
       local behind=${ahead_and_behind[(w)2]}
 
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -174,12 +174,9 @@ git-info() {
     zstyle -s ':zim:git-info:behind' format 'behind_format'
     zstyle -s ':zim:git-info:diverged' format 'diverged_format'
     if [[ -n ${ahead_format} || -n ${behind_format} || -n ${diverged_format} ]]; then
-      # Gets the commit difference counts between local and remote.
-      local ahead_and_behind_cmd='command git rev-list --count --left-right HEAD...@{upstream}'
-
-      # Get ahead and behind counts.
-      local ahead_and_behind      
-      ahead_and_behind=${(z)$(${ahead_and_behind_cmd} 2>/dev/null)}
+      # Get ahead and behind counts from the commit difference counts between local and remote.
+      local ahead_and_behind
+      ahead_and_behind=${(z)$(command git rev-list --count --left-right HEAD...@{upstream} 2>/dev/null)}
       local ahead=${ahead_and_behind[(w)1]}
       local behind=${ahead_and_behind[(w)2]}
 


ericbn:
I see you inlined the `_cmd` variable. When you say "enable proper ahead / behind behavior", what are you trying to fix there?

Tuurlijk:
git-info kept giving me back 0 behind where it should have been 234 behind.

ericbn:
This fixes git-info failing in `zsh 5.0.8` (as reported in #176). Now, what I found is that just breaking the variable assignment as below is enough to fix the issue:
```
--- a/modules/git-info/init.zsh
+++ b/modules/git-info/init.zsh
@@ -178,7 +178,8 @@ git-info() {
       local ahead_and_behind_cmd='command git rev-list --count --left-right HEAD...@{upstream}'

       # Get ahead and behind counts.
-      local ahead_and_behind=$(${(z)ahead_and_behind_cmd} 2>/dev/null)
+      local ahead_and_behind
+      ahead_and_behind=$(${(z)ahead_and_behind_cmd} 2>/dev/null)
       local ahead=${ahead_and_behind[(w)1]}
       local behind=${ahead_and_behind[(w)2]}
```

And I wonder, how come? @Tuurlijk, @Eriner, any clue on why we would need to split var declaration and assignment like that?

Eriner:
@ericbn

> Changes between versions 5.0.8 and 5.1
> The builtins declare, export, local, readonly and typeset now have corresponding reserved words.

This is another one of those classic cases of old versions of zsh behaving in unexpected and unsupported ways. The solution here is to update zsh, not make code changes.

> 5.0.8 was released on May 31, 2015.

Long term solution is for zim to explicitly only support the _latest_ version of zsh. I, or someone else, should PR a change to the README stating this, as it's gotten out of control. At least half of the issues that have arisen are a direct result of those using old versions of zsh, and we (I) have committed many workarounds like above, splitting the variable, and I'm tired of it. I think requiring that users use the latest version of zsh isn't a lot to ask, and will reduce the burden on all of us debugging **stupid** issues like this. I mean, seriously. The prior zsh behavior (that is triggering what is reported here as a bug) was dumb and not friendly, and was subsequently changed over two years ago. If we can't rely on code changes that have been introduced over **two years ago** being present, then we may as well be writing this whole thing in bash.

To continue the rant, this is why supporting anything apple is so bloody annoying. Apple has demonstrated a complete incapability of updating their packages on a reasonable timeline, forcing  people to use brew to install zsh. Not only that, but many peoples are not aware of Apple's sloth, and use the Apple "it just works; we know because we haven't update it in years" provided zsh package, and end up in situations like this, which is even more painful when the user is proactively attempting to contribute to the project @Tuurlijk :+1:. Apple is on par with Debian Stable. Let that sink in.

Tuurlijk:
I would like to use the new version of zsh, but it seems so much slower than the stock osx one. Why is that?

5_0_8-vs-5_3_1

ericbn:
I got a difference of ~0.16s vs. ~0.19s on zsh 5.0.8 vs. 5.3.1, respectively. But I don't think this should be the point here.

@Eriner, what if, instead of supporting only "the latest zsh version", we make a cut and support some specific version onwards? Also, we could make .zim/init.zsh fail with an error message if zsh is not the minimum version. What do you think?

Looks like we're very close to supporting >= 5.0. And we even already have specific code taking care of >= 5.1 or >= 5.2 at [modules/environment/init.zsh](https://github.com/Eriner/zim/blob/fcf060482bbe3f43d97e755530baf26983e09c80/modules/environment/init.zsh) for example.

Eriner:
> what if, instead of supporting only "the latest zsh version", we make a cut and support some specific version onwards?

@ericbn But what would the point of that be? Is there a use case/argument for using a zsh version from a year ago vs the current release? This is an important question, because that would be the only reason to not exclusively support the latest version. I am unable to think of a use case where users will have access (today) to zsh 5.1 but not 5.2, or 5.0 but not 5.3 -- it just doesn't make sense, and unless there is some justification for this, I'm making the call to drop support for non-current versions. If the OS providers packages are out of date, that's not a zim issue, but a distro/packager issue.

ericbn:
> I'm making the call to drop support for non-current versions. If the OS providers packages are out of date, that's not a zim issue, but a distro/packager issue.

@Eriner, you've got a point there! I was looking at the problem in a different way before, but I prefer the way you're putting it. I agree with you.

This will even allow us do remove code that is still taking care of older zsh versions, like the example I gave [before](#issuecomment-302888739).

Do you think it makes sense to make zim/init.zsh fail with an error message if zsh version < 5.3?

Eriner:
> Do you think it makes sense to make zim/init.zsh fail with an error message if zsh version < 5.3?

Yes, I think that will lessen the issue report volume as well.

Eriner:
I would suggest we deprecate support for < 5.3 in the init file now, and in two weeks on June 3rd, we will begin allowing/merging breaking changes for < 5.3. 

The majority of users _should_ already be on a current version of zsh, but if they aren't, two weeks is plenty of time of advance notice. After that, we will `return 1` on zsh != latest minor patch and merge breaking changes. Any issues with this?

lnicola:
CentOS 7 has `zsh 5.0.2`.

ericbn:
@lnicola, please let's continue this discussion at #184.

Issue #175, opened 2017-05-09
Don't overwrite grep colours


nunocastromartins:
`grep` colours seem to be hard coded [here](https://github.com/Eriner/zim/blob/master/modules/utility/init.zsh#L39-L40).

The default colours look bad with my colour scheme.
I may be wrong, but `.zshenv` seems like the _proper_ place to set `GREP_COLORS`.
However, it is sourced before Zim loads, making Zim overwrite my settings.

As a result, I'm currently setting `GREP_COLORS` in `.zshrc`, to overwrite Zim.
Once again, correct me if I'm wrong, but I believe that I shouldn't _have_ to set the variable in `.zshrc`.
Would it be possible for Zim not to set the variable if it is already defined, making it possible to set `GREP_COLORS` in `.zshenv`?

I'm not very fluent in bash, but would something like this do the trick?
```bash
DEFAULT_COLOR='37;45'
export GREP_COLOR=${GREP_COLOR:-"${DEFAULT_COLOR}"}   #BSD
export GREP_COLORS=${GREP_COLORS:-"mt=${GREP_COLOR}"} #GNU
```

Another option would be to allow users to specify the colours in `.zimrc`.

What are your thoughts on this?
Thank you for your work so far!

Tuurlijk:
Same goes for other exports in all modules.

Something like this should fix it for the `utility` module.
```
  if [ -z $GREP_COLOR ]; then
    export GREP_COLOR='37;45'             #BSD
  fi
  if [ -z $GREP_COLORS ]; then
    export GREP_COLORS="mt=${GREP_COLOR}" #GNU
  fi
```

PR #179, opened 2017-05-10
Prevent overwriting of defined environment vars

Tuurlijk:
Env vars set in `.zshenv` should not be overwritten.
However, it is sourced before Zim loads, making Zim overwrite the settings.

Fixes: #175

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -22,10 +22,10 @@ if [[ ! -z ${terminfo[colors]} ]] && (( ${terminfo[colors]} >= 8 )); then
     # BSD
 
     # colours for ls and completion
-    if [[ -z ${LSCOLORS} ]]; then
+    if (( ${+LSCOLORS} )); then
       export LSCOLORS='exfxcxdxbxGxDxabagacad'
     fi
-    if [[ -z ${LS_COLORS} ]]; then
+    if (( ${+LS_COLORS} )); then
       export LS_COLORS='di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:'
     fi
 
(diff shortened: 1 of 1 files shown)


modules/utility/init.zsh
```diff
@@ -22,8 +22,12 @@ if [[ ! -z ${terminfo[colors]} ]] && (( ${terminfo[colors]} >= 8 )); then
     # BSD
 
     # colours for ls and completion
-    export LSCOLORS='exfxcxdxbxGxDxabagacad'
-    export LS_COLORS='di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:'
+    if [[ -z ${LSCOLORS} ]]; then
```
Eriner: I'd like to see this style of these checks replaced with `(( ${+LSCOLORS} ))`. If everyone was using zsh 5.3+, we could use the nice `[[ -v LSCOLORS ]]` syntax, but alas.

Eriner:
Thanks for the PRs! :+1:

modules/utility/init.zsh
Tuurlijk: Can do, but here the `-z` option is used as in: https://github.com/Eriner/zim/blob/master/modules/environment/init.zsh#L42

Tuurlijk: It's working here on zsh 5.0.8 (x86_64-apple-darwin15.0)

Eriner: Hm... 76ff1bd was 6 months ago, 5.3 released Dec 11, 2016. Well, I personally am okay with only supporting the _latest_ version of zsh, as it makes our job easier. This means we may get complains from users about things not working, however in many cases means they're using some ancient mac-default version of zsh from 2011, the solution of "update your shit" is the same, so I'm actually okay with this.

`-z` it is.

Eriner: Er, I misspoke and confused `-z` with `-v`. It either needs to be `(( ${+LSCOLORS} ))` or `-v`, and the reference you gave should be changed.

Tuurlijk:
Fixed the utility file. Other `[[ -z ${BLAH} ]]` occurrences will be fixed in a separate pr.

Tuurlijk:
Wait . . . isn't `[[ -z ${BLAH} ]]` different from  `(( ${+BLAH} ))`?

I think we need the conditions to be: `if ! (( ${+BLAH} )); then`

ericbn:
Instead of `! (( ${+BLAH} ))`, I recommend `[[ -z ${BLAH} ]]`, which checks if the string is not empty. The first should be primarily used for arithmetic expressions.

Eriner:
> Instead of ! (( ${+BLAH} )), I recommend [[ -z ${BLAH} ]], which checks if the string is not empty. The first should be primarily used for arithmetic expressions.

See my comment in #180

PR #180, opened 2017-05-10
Use zsh style tests for empty vars

Tuurlijk:

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -39,7 +39,7 @@ unsetopt HUP
 unsetopt CHECK_JOBS
 
 # Set less or more as the default pager.
-if [[ -z ${PAGER} ]]; then
+if (( ${+PAGER} )); then
   if (( ${+commands[less]} )); then
     export PAGER=less
   else
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -39,7 +39,7 @@ unsetopt HUP
 unsetopt CHECK_JOBS
 
 # Set less or more as the default pager.
-if ! (( ${+PAGER} )); then
+if (( ! ${+PAGER} )); then
   if (( ${+commands[less]} )); then
     export PAGER=less
   else
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -39,7 +39,7 @@ unsetopt HUP
 unsetopt CHECK_JOBS
 
 # Set less or more as the default pager.
-if (( ${+PAGER} )); then
+if ! (( ${+PAGER} )); then
   if (( ${+commands[less]} )); then
     export PAGER=less
   else
--- a/modules/fasd/functions/fasd
+++ b/modules/fasd/functions/fasd
@@ -41,20 +41,20 @@ case $1 in
         fi
 
         # set default options
-        ! (( "${+_FASD_DATA}" )) && _FASD_DATA="$HOME/.fasd"
-        ! (( "${+_FASD_BLACKLIST}" )) && _FASD_BLACKLIST="--help"
-        ! (( "${+_FASD_SHIFT}" )) && _FASD_SHIFT="sudo busybox"
-        ! (( "${+_FASD_IGNORE}" )) && _FASD_IGNORE="fasd ls echo"
-        ! (( "${+_FASD_SINK}" )) && _FASD_SINK=/dev/null
-        ! (( "${+_FASD_TRACK_PWD}" )) && _FASD_TRACK_PWD=1
-        ! (( "${+_FASD_MAX}" )) && _FASD_MAX=2000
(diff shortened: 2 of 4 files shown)


ericbn:
How is `! (( ${+BLAH} ))` better than `[[ -z ${BLAH} ]]`? I see you're testing if `${BLAH}` exists with `${+BLAH}`, but I think it does not read as better as the original code...

ericbn:
Good job on finding all those occurrences, though!  👍

Eriner:
> How is `(( ! ${+BLAH} ))` better than `[[ -z ${BLAH} ]]`

Note: `(( ! ${+BLAH} ))` not `! (( ${+BLAH} ))`

The reason for this is that this style will catch set but empty variables. For example:

Testing if variable is set with math expression, only evaluates to true if the variable is truly unset.
```
$ foo='exists'
$ if (( ! ${+foo} )); then print "it's unset"; fi    # doesn't print anything, as the variable exists
$ foo=''
$ if (( ! ${+foo} )); then print "it's unset"; fi    # also doesn't print anything, as the variable exists even though it's an unset/blank string
$ unset foo
$ if (( ! ${+foo} )); then print "it's unset"; fi 
it's unset
```

Testing if variable is set with `-z` test, evaluates to true if variable is unset **or** an empty string.
```
$ foo='exists'
$ if [[ -z $foo ]]; then print "it's unset"; fi    # doesn't print anything, as the variable exists
$ foo=''
$ if [[ -z $foo ]]; then print "it's unset"; fi    # triggers on a blank string!!
it's unset
$ unset foo
$ if [[ -z $foo ]]; then print "it's unset"; fi    # triggers on unset variable.
it's unset
```

Like I mentioned elsewhere, if we decide to only support the latest version of zsh (which I think ultimately is a good idea), we can use the `[[ -v varname ]]` syntax...

ericbn:
These changes might introduce undesired behavior like in:

    url="$(command git config --file "$(git-root)/.gitmodules" --get "submodule.${src}.url")"
  
    if ! (( "${+url}" )); then # instead of if [[ -z "${url}" ]]; then

The above code will always evaluate the new test to true, since `url` will always be defined at that point. But `url` will be non-empty only if the command executed before succeeded.

Also, why would we avoid overriding an empty value of `PAGER` by a a value of `less`, for example? 

And even the existing code:

    if (($+ztermtitle)); then

Does it make sense to do

    precmd() { print -Pn "\e]0;${ztermtitle}\a" }

if `${ztermtitle}` is an empty string?

I think blindly replacing all `[[ -z ${BLAH} ]]` by `(( ! ${+BLAH} ))` is dangerous, as per my first example.

And I would actually go the other way around and also replace all `(( ${+BLAH} ))` by `[[ -n ${BLAH} ]]`. I cannot think of a situation where it makes sense to check if a variable exists to later use it's value, even when the value is empty... Also, checking if a variable does not exist to override it's value has less effect than checking if the variable in empty...

Tuurlijk:
I see `[[ -z ${BLAH} ]]` all the time in documentation and other scripts. So I'm interested in finding the point in time where it was introduces. That way we can decide if it's OK to leave it in or not.

It also sounds like we need a test framework, so we can avoid accidentally breaking the config. We can also build in speed tests.

And the choices we make on usage of ways to check for vars should be written down in the WIKI. How can I contribute there?

Eriner:
> I cannot think of a situation where it makes sense to check if a variable exists to later use it's value, even when the value is empty... Also, checking if a variable does not exist to override it's value has less effect than checking if the variable in empty...

If the variable has been set, even if it's a blank string, it means that something other than zim has run into that variable, be it user configuration or otherwise. @ericbn the code example you gave is of course correct, and there are cases where you _do_ want to explicitly check if the variable is a blank string. However, in cases where we are providing "better defaults", it makes sense to me to not set variables that have been set before the zim code hits it; in other words, if `PAGER` exists by that point, don't do anything as it's been set outside of zim. Also in cases where it's appropriate to use, `(( ${+var} ))` syntax is faster, as it's not evaluating the contents of the variable and merely getting a Boolean as to if it's set or not.

ericbn:
@Eriner, sounds good using `(( ${+var} ))` for scenarios where we're "providing 'better defaults'" (using the negated `(( ! ${+var} ))` form), or other cases where it makes sense to just check if var exists, regardless if it's empty or not...  👍 

@Tuurlijk, could you please review your changes based on this?

Tuurlijk:
@ericbn Done

ericbn:
@Tuurlijk, I still see the "breaking" change in git-submodule-move. Also, I'm not sure about all the changes in the fasd module, it they're not breaking anything... The first few changes look good.

Tuurlijk:
@ericbn Ok, reverted the changes you're unsure about.

Issue #94, opened 2016-10-13
merged file of template file and user file raise parse error


ghreyt:
I think it needs to put newline between **template_file** and **user_file**
in installation part in **README.md**.

``` shell
setopt EXTENDED_GLOB
for template_file ( ${ZDOTDIR:-${HOME}}/.zim/templates/* ); do
  user_file="${ZDOTDIR:-${HOME}}/.${template_file:t}"
  touch ${user_file}
  ( print -rn "$(<${template_file})$(<${user_file})" >! ${user_file} ) 2>/dev/null
done
```

My initial **.zshrc** was like

``` shell
# comment about something
```

and installation loop above produces the **.zshrc** like

``` shell
... zim zshrc
...
# Source zim
if [[ -s ${ZDOTDIR:-${HOME}}/.zim/init.zsh ]]; then
  source ${ZDOTDIR:-${HOME}}/.zim/init.zsh
f# comment about something
```

, which raise parse error.

So I think it might be better to change printing line to put new line between template and user file
like below (or any better way).

``` shell
  ( print -rn "$(<${template_file})"$'\n'"$(<${user_file})" >! ${user_file} ) 2>/dev/null
```

Eriner:
Yeah, that "script" blob is very finicky and prone to error. I wrote it to try to make installation easy, but without telling people to `curl xyz | zsh`, because I wanted the process to be transparent. With the move to the new organization and project structuring, this will be changed. I'll leave this issue open until I've finished that. Thanks for the report :+1:

kaznovac:
#280 seems related

PR #182, opened 2017-05-20
Install via script instead of 5 manual steps

foldback:
Basically what it says on the tin can. Allows for `sh -c "$(curl -fsSL --proto-redir -all,https https://raw.githubusercontent.com/Eriner/zim/master/tools/zim_install)"`, which makes installs both safer and easier. See individual commits for the why and how.
Its been fully tested and is in the zim code style :)


Eriner:
@jorvi The work here looks good, though I'll have to review the script more thoroughly and do a few test-cases. Ideally I'd work out a CI system for this and zim in general, but that's a lot of work into, AFAIK, uncharted territory. I'll have to spend some time going over the script and make some stylistic comments. Hopefully I can get the easy ones back to you later this evening or tomorrow. Thanks!! :+1:

Eriner:
+cc @ericbn, might be interested in this one as well.

ericbn:
I like the idea of having a script to install Zim. Very good initiative, @jorvi! I see you worked a lot there. Is this based on any other install script or you did everything from scratch?

I haven't tried running the script yet, but I've read the whole code. I'm making a list of what the script does, to make the discussion around the code easier:
1. Check if zsh is installed with `$(grep /zsh$ /etc/shells | wc -l)`, and abort if not installed.
2. Check if zim is already installed with `[ -d "${ZIM_DIR}" ]`, where `ZIM_DIR="${ZDOTDIR:-${HOME}}/.zim"`, and abort if already installed.
3. Check if other frameworks exist with `[ -d "${HOME}/.xyz" ]`, where xyz is `oh-my-zsh`, `zplug` and `zprezto`, and do `mv "${HOME}/.xyz" "${HOME}/.xyz_zimbackup` if any exists.
4. Check if git is installed with `hash git >/dev/null 2>&1 ||`, and abort if not installed.
5. Check if msysgit is installed with `"${OSTYPE}" = cygwin` and `git --version | grep msysgit`, and abort if installed.
6. Clone the zim repo.
7. Backup user files with `mv "${HOME}/.xyz" "${HOME}/.xyz_zimbackup"`, where xyz is `zshrc` and `zlogin`.
8. Copy template files to `${HOME}`.
9. If current shell is not zsh, attempt to "auto-change" the user shell to zsh, where `CURRENT_SHELL=$(expr "${SHELL}" : '.*/\(.*\)')`, zsh path is computed with `chsh -s $(grep /zsh$ /etc/shells | tail -1)` and chsh existence is checked with `hash chsh`.

I have a few thoughts here:
- Why the changes in the template files?
- And why aren't the template files being appended to the user ones, as per the current instructions?
- I would make the script fail if zsh is not the current shell.
- I would maybe make the script itself a zsh script, and then check for other commands with `(( ${+commands[git]} ))`, for example.
- Why move the other frameworks to backup directories? If they're not being sourced anymore, their directories are just innocuous.

Tmplt:
While it's all very convenient, should `curl | sh`ing really be endorsed? Is the common distaste against the practise only aimed against URLs such as `install.example.tld`, which doesn't contain `/path/to/install_script`, or does the convenience simply outweigh the possible security issues in this case?

ericbn:
@Tmplt, I remember this was discussed before, and the approach was to only accept [signed commits to the install script](https://github.com/zimframework/zim/issues/1#issuecomment-250949856).

Eriner:
Yes, I'll sign any commits that touch the `curl`'d file. While the risk is the same between all of the files (as they will inevitably be sourced), it ensures that I 'sign-off' on that critical code bit. Additionally, providing strong curl/wget options mitigates additional problems.

Tmplt:
That's a neat solution. But wouldn't this require the user to import the key and verify the script? It can't verify itself, can it?

Eriner:
> But wouldn't this require the user to import the key and verify the script? It can't verify itself, can it?

Since I'll be the only one signing them, we should just be able to verify against my static long id; should be able to build it into the one-liner, I just haven't done it before so I'd have to look into how to walk the ref chain to verify all commits to a particular file match a given id.

jorvi:
@ericbn 
Word of warning: I can sound  a bit abrasive at times, but its all meant in good faith :)

* Why the changes in the template files?
 * `# # # old settings go here # # #` is just a pasting guide so people know where in the file they should past their old stuff. Most users don't know what an interactive shell is, nor if their old stuff should go before or after `$SOURCE_ZIM`. 
I had 2-3 times where the old way (prepend) actually pasted in the middle of my `.zshrc` lines. Try it now: use a `.zshrc` with a commented line at the top, and a `if $X; then whatever; fi;` conditional on the lines under that , then install zim: you'll end up with a mangled `.zshrc`. That's why I'd rather `cp` the templates into place and let the user take care of business. Actually, I just had a realization\*..
The other changes are mainly cosmetic.. making variables uniform (use `$VARIABLE` everywhere, instead of `$VARIABLE` in one place and `$variable` in another, and properly close every line of a `if then; 
  stuff; 
The renames are so I can do a simple `cp -a` instead of having to rename the files in script.
fi;` conditional)
* I would make the script fail if zsh is not the current shell.
 * Good idea
* I would maybe make the script itself a zsh script, and then check for other commands with (( ${+commands[git]} )), for example.
 * Cool idea :) I'm not terribly well-versed in zsh-isms yet though..
* Why move the other frameworks to backup directories? If they're not being sourced anymore, their directories are just innocuous.
 * They are, but there's a fat chance some users will blindly paste their old config into Zim's template `.zshrc`, and then they're actually sourcing two frameworks. We prevent that with the other frameworks moved to backup directories.

The general idea, for me, with this script is  to make the install via this script the most fast, safe and convenient way to get Zim.  If `git clone && checksig && ./install.sh` is much safer than `curl | sh`, that's fine as its still one copy-paste-enter for the user. But, for example, sudo should be avoided if at all possible.

\* We can use something like `sed` to replace `# # # old settings go here # # #` with the contents of `.zshrc_zimbackup`. We could even check for (and cut out) `source $FRAMEWORK`!

Also, if we really wanna we could: 
`if not script running in zsh
NO_ZSH=print "Error: please manually install and switch to zsh!'
    if not zsh exists
        if not homebrew installed
            cp or curl zsh binary\*
            changeshell to zsh
        elif homebrew installed
            brew install zsh
            changeshell to zsh
        else
           $NO_ZSH and exit
        fi
      $NO_ZSH and exit 
    elif zsh exists
        changeshell to zsh
    else
        $NO_ZSH
    fi
unset NO_ZSH
fi

\* cp or curl to /usr/local/bin. We can either include the up-to-date zsh binary in the repo (which is under our control) or `curl https` it in from sourceforge.net

Eriner:
@jorvi 

> We prevent that with the other frameworks moved to backup directories.

I actually don't mind this, as long as we print a message in a `trap` function to let the user know that all their stuff has been moved. The current 5-liner can do some weird parsing things, as I know some of you here have experienced; it was the best I could do while attempting to make the whole process as transparent as possible! :)

> `git clone && checksig && ./install.sh`

I think this will ultimately look something like this:
```
git clone https://github.com/Eriner/zim.git ~/.zim && \
git do some gpg commit magic && \
zsh ~/.zim/tools/install.zsh
```
It will never be as condensed as `curl x | sh`, but it sure won't be the blob it is now.

> use `$VARIABLE` everywhere, instead of `$variable` in one place and `$VARIABLE` in another

In the .zlogin, I had done this explicitly and with purpose. Because everything is run in a sub-shell (notice the `(){` that wraps everything), the variables are lost to the breeze when in concludes, so we don't have to even `local` them, or worry about collision (hence why we would otherwise use use caps).

> Since its the same as `if X; then stuff; fi;`, and you wouldn't leave stuffunclosed.

Actually, the only reason to use the `;` at all is because in order to (where appropriate) fit the conditional on a one-liner, we need to explicitly end the statement. Where we don't have to cram the conditional into a single line, the line separations themselves are implicit and thus we don't need the `;`. While being explicit is always good, this doesn't improve code readability IMO.

> But, for example, sudo should be avoided if at all possible.

The install is, and always will be, per-user. I don't plan on explicitly supporting `/usr/` installs, thus we don't and should never need sudo for installation.

> `# # # old settings go here # # #`

Perhaps "personal configuration below this line" or something similar would be more clear? zim should be the first item to be sourced, and all additional user configuration should go after it. This way, even if we step on some variables, the user's configuration should overwrite it. (Though, some work has been done in past PRs to resolve some of this issue, hopefully it won't remain this way forever and we won't have to care about when/where we're loaded).

jorvi:
* No longer close every line with `;`
* Reduce amount of variables by doing the magic inside `[ ]`
* Variables all in lowercase since all created inside subshell
* Attempt to install `zsh` via `homebrew` if `zsh` doesn't exist on the system
* Fix some typos
* Restore (most of) the `.template` files to former state
* Adjust `README.MD` to install with `git clone && install.sh` instead of using `curl`

I've also been pondering verification. `Shasum` or `MD5` seems like a nice stopgap until we figure out how to GPG verify. Something like `if [ ! "$(shasum "${zim_dir}/tools/zim_install")" = "$shavalue" ]; then ...`.  I've also figured out that its possible to curl a page (say `https://github.com/jorvi/zim/commits/master/tools/zim_install`) and ` | egrep -w 'Verified|jorvi'`. If we could somehow just `curl` the `class="commit-links-cell table-list-cell"` on that page, we could just check if the latest commit contains both 'Verified' and 'Eriner'.

Eriner:
@jorvi good on the changes.

> The end result is that even on a minimal system that only has `sh` Zim gets itself up and running

The only issue with this is that it's slightly out-of-scope for the project. What I mean by that is, while pulling it from homebrew if $OS == Mac is fine and all, but what about Arch? What about Ubuntu? What about Fedora? What about Alpine? The issue I have with this approach is that it's platform-dependent, and may not be what the user wants. I'm totally okay with the script targeting `/bin/sh` and prompting to get zsh if not installed/sufficient version, but we'd have to account for all the reasonable systems, or dump `echo "zsh is currently not installed" && echo "unable to detect operating system for assisted installation"`. If we can do those things, :+1: to the idea.

jorvi:
@Eriner 
Having the zsh binary, manpages and PRCE dependency inside the repo (in /zsh) and just copying/symlinking them to `/usr/local/bin` is out of the question? It'd be a bit of a hassle since we'd need to keep those binaries up-to-date, but it would also mean you can guarantee that zsh is available on the system. I can totally understand telling the user that its their problem though, none of the other frameworks (OMZ, Prezto, Zplug) include zsh or a zsh loader.

Eriner:
@jorvi Yeah unfortunately I think this would introduce too much complexity both from a zim management standpoint, and from an admin/user standpoint. For instance, if we take change of managing the zsh binary, what happens if the user explicitly installs zsh through their package manager? Essentially, we'd have to pray to the PATH gods that we get loaded first. For example, if we manage zsh and update the package in `/usr/local/bin` when needed, if we ever just run `zsh -version` and the binary is called from `/bin/zsh` and returns an old version, we might end up upgrading our managed copy of the binary on every check of the version. Lots of weird collisions that could happen based on the user's environment that's out of our control. And since that is the case, I'd prefer to just use the system's package manager (brew for OSX) if we implement this functionality fully. If there are issues, we can just print an error and `return 1`, or prompt the user for the script to attempt to resolve the issue (install zsh). Thoughts?

Eriner:
I think targeting `/bin/sh` is a good idea, checking for `zsh` is a good idea, and attempting to install `zsh` (after prompting and notifying user) is a good idea; it just has to be OS dependent and `return 1` with a generic "zsh not found; please install and re-run script" if we can't detect/target the current ENV OS.

jorvi:
@Eriner can't target everything though. macOS (brew), Ubuntu (apt or apt-get), Fedora (rpm) and Arch (pac) should be enough, I think? That should cover 99% of the Unix crowd. I'll get going..

jorvi:
@Eriner 

I had a realization during the building of the zsh_check+_install: all the systems we would do this on (macOS, Debian/Ubuntu, Fedora, Arch) already come with zsh built in. There's no need to run a check, because zsh is pretty much guaranteed to be there. Its needless complexity.
Version checking is possible but is a pain in the butt (you need to use `bc` or `awk` because sh/bash/zsh only support integer math).

Anyway, what other work on the installer (besides stripping out the `brew install zsh`) needs to be done before its gonna be merged?

ericbn:
@jorvi, following the lines of #184, I would make the install script fail if the zsh version is < 5.3.

jorvi:
@ericbn but macOS Sierra (don't know about High Sierra) has ZSH 5.2.. its pretty bad for ease of use if .zim _requires_ the majority of its users to either use brew or manually manage a zsh binary..

Eriner:
@jorvi zsh has `is-at-least`, so version checking is easy.

jorvi:
@Eriner @ericbn 
Updated :) checks if zsh is not on the system or if the zsh version is not at least 5.2. If either of these cases is true, it tries to install via `brew`, `apt-get`, `dnf` or `Pacman`. This should cover about 99% of Zim users.

jorvi:
@Eriner @ericbn ping

Eriner:
Now that #184 is closed, I'll take a more careful look at this.

tools/zim_install
```diff
@@ -0,0 +1,158 @@
...
+    printf "${yellow}Please inspect ${HOME}/.zshrc_zimbackup and copy over${normal}\n"
+    printf "${yellow}wanted settings to ${HOME}/.zshrc${normal}\n"
+  fi
+  if [ -f "${HOME}/.zlogin_zimbackup" ] || [ -L "${HOME}/.zlogin_zimbackup" ]; then
+    printf "${yellow}Please inspect ${HOME}/.zlogin_zimbackup and copy over${normal}\n"
+    printf "${yellow}wanted settings to ${HOME}/.zlogin${normal}\n"
+  fi
+
+  # Load zsh
+  env zsh
```
Eriner: Maybe instead of `env`, which launches a new sub-process (and requires you `exit` twice), we use `exec`.

jorvi: Does `exec` check `path` so we load the correct zsh? with `env` we can be sure we use `/usr/local/bin/zsh` if its installed and `/bin/zsh` otherwise.

Eriner:
@jorvi I have some time off work coming up in about two weeks and would like to give this project a bit of love, so I'll be reviewing your PR more fully then. Sorry for the delays, that's just how things go sometimes :)

If @ericbn has some time, I'd like his opinion as well, but I'm sure he's busy too. He's also been doing a tremendous job helping to review things :+1:

tools/zim_install
Eriner: > exec [ -cl ] [ -a argv0 ] [ command [ arg ... ] ]
>
> Replace the current shell with an external command **rather than forking**. With **-c clear the environment**; with -l prepend - to the argv[0] string of the command executed (to simulate a login shell); with -a argv0 set the argv[0] string of the command executed. See Precommand Modifiers.

Eriner: That would lead me to believe that just calling the `exec` builtin should be fine, but of course, that will need to be tested. I don't see why it wouldn't work, though.

Eriner:
Closing this should close #94; we're now telling users how to install the templates instead of a heavy-handed glob.

tools/zim_install
```diff
@@ -0,0 +1,158 @@
...
+    if printf "${answer}\n" | grep -iq "^y"; then
+        return 0
+    else
+        return 1
+    fi;
+  }
+
+  # Check if `zsh` exists, or if `zsh` isn't too old
+  if [ ! "$(grep "/zsh$" "/etc/shells")" ] || \
+    env zsh -c "autoload -Uz 'is-at-least'; ! is-at-least 5.2"; then
```
ericbn: Shouldn't we use `&&` instead of the first `;`?

tools/zim_install
```diff
@@ -0,0 +1,158 @@
...
+
+  # Yes or no function
+  yesno() {
+    printf "${*} "
+    read answer
+    if printf "${answer}\n" | grep -iq "^y"; then
+        return 0
+    else
+        return 1
+    fi;
```
ericbn: No need for this `;` at the end

Eriner:
@jorvi Would you like to address the comments @ericbn and I have made, or would you like me to take care of it at merge?

tools/zim_install
jorvi: Feel free to remove it, its a personal style thing I guess :)

tools/zim_install
jorvi: How do you figure that? It's a basic `if x | y; then ..`

jorvi:
@Eriner I don't have access to a proper laptop for the next week and a half, so feel free to merge/edit however you and @ericbn like :)

partounian:
Is there help needed here? I'm on a open source crusade fixing/improving stuff that's cool.

ericbn:
Hey @partounian. Welcome crusader! I'm glad you found zim during your cool-stuff path...  :- )

Yeah, first this needs to be rebased to master and conflicts fixed, then there are the latest comments from @Eriner and me to be taken into account...

partounian:
Funny enough installing Zim didn't work for me, so I'll have to debug that over the weekend. (Got errors for writing zwc files to home)

Eriner:
When this issue is finally wrapped up (I accept most of the blame here, I've been neglecting this project and shafting @ericbn with a lot of maintenance work, sorry @ericbn!), I think we should tag v1.0.

AtomicCoding:
@ericbn @Eriner 
Just made some of my own changes, made some fixes for the `yesno` function and for package installation, removed the need for adding a dot before the templates by using file globbing and a for loop, along with cloning the git in the installation function. I feel we should stop using returns as they keep closing bash even though they're in a function.
Enough talk, here's the code:
```zsh
main() {
  # Use colors, but only if connected to a terminal, and that terminal
  # supports them.
  if which tput >/dev/null 2>&1; then
    ncolors=$(tput colors)
  fi
  if [ -t 1 ] && [ -n "${ncolors}" ] && [ "${ncolors}" -ge 8 ]; then
    red="$(tput setaf 1)"
    green="$(tput setaf 2)"
    yellow="$(tput setaf 3)"
    blue="$(tput setaf 4)"
    bold="$(tput bold)"
    normal="$(tput sgr0)"
  else
    red=""
    green=""
    yellow=""
    blue=""
    bold=""
    normal=""
  fi

  # Only enable exit-on-error after the non-critical colorization stuff,
  # which may fail on systems lacking tput or terminfo
  set -e

  # Yes or no function
  yesno() {
    printf "${*} (y or n) >>> "
    read answer
    echo "$answer"
    if [ "$answer" = "y" ]; then
        echo "YES"
    else
       false
    fi
  }

  # Check if `zsh` exists, or if `zsh` isn't too old
    if $(env zsh -c "autoload -Uz 'is-at-least' && is-at-least 5.2") || $(command -v "zsh" > /dev/null); then
    true
  else
    yesno "Do you want the installer to automatically install 'zsh'?"
    echo "installing"
    # try installing from 'brew' (Mac)
    if command -v "brew" 2>/dev/null; then
      brew install "zsh" >/dev/null 2>&1
      printf "${yellow}Installing zsh via Homebrew.. ${normal}\n"
    # try installing from 'apt-get' (Debian/Ubuntu)
    elif command -v "apt" 2>/dev/null; then
      echo "Installing..."
      sudo apt install "zsh"
    # try installing from 'dnf' (Fedora)
    elif command -v "dnf" 2>/dev/null; then
      dnf install "zsh"
    # try installing from 'pacman' (Arch)
    elif command -v "pacman" 2>/dev/null; then
      pacman -S "zsh"
    else
      printf "${red}Error: 'zsh' isn't available.. ${blue}please install manually!${normal}\n"
      return 1
    fi
  fi

  # Check if Zim is already installed
  # .Zim causes more problems than .zimrc
  if [ -e "${ZDOTDIR:-${HOME}}/.zim" ]; then
    printf "${red}Error: You already have Zim installed!${normal}\n";
    return 1
  fi

  # Check if other frameworks are installed
  if [ -d "${HOME}/.oh-my-zsh" ]; then
    mv "${HOME}/.oh-my-zsh" "${HOME}/.oh-my-zsh_zimbackup"
    printf "${yellow}Oh-My-Zsh detected and backed up to\
     ${HOME}/.oh-my-zsh_zimbackup${normal}\n"
    printf "${blue}This is done to prevent framework conflicts.${normal}\n"
  elif [ -d "${HOME}/.zplug" ]; then
    mv "${HOME}/.zplug" "${HOME}/.zplug_zimbackup"
    printf "${yellow}Zplug detected and backed up to\
     ${HOME}/.zplug_zimbackup${normal}\n"
    printf "${blue}This is done to prevent framework conflicts.${normal}\n"
  elif [ -d "${HOME}/.zprezto" ]; then
    mv "${HOME}/.zprezto" "${HOME}/.zprezto_zimbackup"
    printf "${yellow}Prezto detected and backed up to\
     ${HOME}/.zprezto_zimbackup${normal}\n"
    printf "${blue}This is done to prevent framework conflicts.${normal}\n"
  fi

  # Prevent the cloned repository from having insecure permissions.
  umask g-w,o-w

  # The Windows (MSYS) Git is not compatible with normal use on cygwin
  if [ "${OSTYPE}" = "cygwin" ]; then
    if git --version | grep msysgit >/dev/null; then
      printf "${red}Error: Windows/MSYS Git is not supported on Cygwin${normal}\n"
      printf "${red}Error: Make sure the Cygwin git package is installed and is first on the path${normal}\n"
      return 1
    fi
  fi

  # Back-up old zsh dotfiles
  if [ -e "${HOME}/.zshrc" ]; then
    printf "${yellow}Found ${HOME}/.zshrc! ${blue}Backing up to ${HOME}/.zshrc_zimbackup.${normal}\n"
    mv "${HOME}/.zshrc" "${HOME}/.zshrc_zimbackup"
  fi
  if [ -e "${HOME}/.zlogin" ]; then
    printf "${yellow}Found ${HOME}/.zlogin!$ ${blue}Backing up to ${HOME}/.zlogin_zimbackup${normal}\n"
    mv "${HOME}/.zlogin" "${HOME}/.zlogin_zimbackup"
  fi
  if command -v git > /dev/null; then
    printf "${blue}Downloading Zim...${normal}\n"
    git clone --recursive https://github.com/zimfw/zimfw.git ${ZDOTDIR:-${HOME}}/.zim > /dev/null
  else
    printf "${red}Please install git!${normal}\n"
  fi
  # Place template files in ~/
  for file in ${HOME}/.zim/templates/*; do
    cp "$file" "${HOME}/.${file##*/}"
  done

  # If this user's login shell is not already "zsh", attempt to switch
  if [ ! "$(expr "${SHELL}" : '.*/\(.*\)')" = "zsh" ]; then
    # Auto-change shell if platform provides "chsh" command
    if hash chsh >/dev/null 2>&1; then
      printf "${blue}Changing your default shell to zsh!${normal}\n"
      chsh -s $(grep "/zsh$" "/etc/shells" | tail -1)
    # If not, suggest user do so manually.
    else
      printf "${yellow}I can't change your shell automatically because this system does not have chsh.${normal}\n"
      printf "${yellow}Please manually change your default shell to zsh!${normal}\n"
    fi
  fi

  Z="${red}Z${green}"
  printf "${green}               GG            8L8                                                   ${normal}\n"
  printf "${green}              8C08           0CG                                                   ${normal}\n"
  printf "${green}              CC              8f8                                                  ${normal}\n"
  printf "${green}             8t8               fC                                                  ${normal}\n"
  printf "${green}             Ct                018                                                 ${normal}\n"
  printf "${green}            8iG       888       iC                                                 ${normal}\n"
  printf "${green}         88CiG0GGCCCCCCCCCCGGG0ft0888                                              ${normal}\n"
  printf "${green}        0LLLLLLLLLLLLLLLLLLLLLLLLLLLLG                                             ${normal}\n"
  printf "${green}        8LLLLLLLLLLLLLLLLLLLLLLLLLLLLG                                             ${normal}\n"
  printf "${green}        CLLLLLL0880LLLLLLC080CLLLLLL0     8888888888      8                        ${normal}\n"
  printf "${green}        GLLLLL0    0LLLLC8   8CLLLLC     8LLLLLLLLLLt18  0t0                       ${normal}\n"
  printf "${green}        8LLLLC8    8CLLL0     0LLLLG               CtG    8                        ${normal}\n"
  printf "${green}         CLLLC      CLLL0     0LLLL0             8ff8    8G    G 8GCLCCG  8GCCLCG8 ${normal}\n"
  printf "${green}         GLLLL8    8LLLLG     GLLLL8            0tC      0i8  8iffG0880LtLfC0880Ct0${normal}\n"
  printf "${green}         0LLLLC8  8CLLLLL0   0LLLLC            Ct0       0i8  8;f       1t8      Lt${normal}\n"
  printf "${green}         8CLLLLCGGCLLLLLLLCGCLLLLLG          8ff8        0i8  8i0       fC       C1${normal}\n"
  printf "${green}          CLLLLLLLLLLLLLLLLLLLLLLL8         GtC          0i8  8i0       fC       C1${normal}\n"
  printf "${green}          0LLLLLLLLLLLLLLLLLLLLLLC8        Lt0           0i8  8i0       fC       C1${normal}\n"
  printf "${green}          8LLLLLLLLLLLLLLLLLLLLLLG       0tf08888888888  0i8  8i0       fC       C1${normal}\n"
  printf "${green}           CLLLLLLLLLLLLLLLLLLLLL0       CtLLLLLLLLLLLG  8L    L8       C0       0L${normal}\n"
  printf "${green}           GLLLLLLLLLLLLLLLLLLLLL8                                                 ${normal}\n"
  printf "${green}           8GCCCCLLLLLLLLLLLCCCGG                                                  ${normal}\n"
  printf "${green}              88800GGGGG00888                                                      ${normal}\n"
  unset Z
  # For archival purposes
  #   ______      _       _____  __  __                                    _
  #  |___  /     | |     |_   _||  \/  |                                  | |
  #     / /  ___ | |__     | |  | \  / | _ __   _ __  ___ __   __ ___   __| |
  #    / /  / __|| '_ \    | |  | |\/| || '_ \ | '__|/ _ \\ \ / // _ \ / _` |
  #   / /__ \__ \| | | |  _| |_ | |  | || |_) || |  | (_) |\ V /|  __/| (_| |
  #  /_____||___/|_| |_| |_____||_|  |_|| .__/ |_|   \___/  \_/  \___| \__,_|
  #                                      | |
  #                                      |_|

  # Remind user to copy back old settings
  if [ -f "${HOME}/.zshrc_zimbackup" ] || [ -L "${HOME}/.zshrc_zimbackup" ]; then
    printf "${yellow}Please inspect ${HOME}/.zshrc_zimbackup and copy over${normal}\n"
    printf "${yellow}wanted settings to ${HOME}/.zshrc${normal}\n"
  fi
  if [ -f "${HOME}/.zlogin_zimbackup" ] || [ -L "${HOME}/.zlogin_zimbackup" ]; then
    printf "${yellow}Please inspect ${HOME}/.zlogin_zimbackup and copy over${normal}\n"
    printf "${yellow}wanted settings to ${HOME}/.zlogin${normal}\n"
  fi

  # Load zsh
  exec zsh
}

main
```

PR #189, opened 2017-06-07
[history-substring-search] Remove redundant defaults

ericbn:
for options already defined, with the same values, in `zsh-history-substring-search.zsh`. See https://github.com/zsh-users/zsh-history-substring-search/blob/be0fe1fca94c2eea3c3798de3c36acfcebf16943/zsh-history-substring-search.zsh#L46-L48


Eriner:
@ericbn LGTM

PR #191, opened 2017-06-11
init.zsh: Allow arbitrary ZIM location

m45t3r:
This change allow arbitrary ZIM location path by setting a new environment variable, `ZIM_PATH`. If the user does not set it, ZIM falls back to the old `"${ZDOTDIR:-${HOME}}/.zim"` location.

The object of this change is to make ZIM location more flexible and also helps with development (you can set this variable to a development version of ZIM to local test it).

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -3,7 +3,9 @@
 #
 
 # Define zim location
-ZIM="${ZDOTDIR:-${HOME}}/.zim"
+if [[ -z "${ZIM_PATH}" ]]; then
+  ZIM_PATH="${ZDOTDIR:-${HOME}}/.zim"
+fi
 
 # Source user configuration
 if [[ -s "${ZDOTDIR:-${HOME}}/.zimrc" ]]; then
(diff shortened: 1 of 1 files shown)


Eriner:
@m45t3r the zmanage meta module functions would need to change as well.

ericbn:
@m45t3r, why not `ZIM_HOME` instead of `ZIM_PATH`?

m45t3r:
@Eriner Can you give an example? Looking just at the code of `zmanage` I didn't find anything.

@ericbn No reason really, I can change.

Plloi:
Why replace `ZIM`?
If you just set that with the new method everything should be fine.

ericbn:
@m45t3r, the zmanage meta module functions are in the `tools` subdirectory. You'll also want to touch `templates/zshrc` and `templates/zlogin`.

m45t3r:
For some reason GitHub is not detecting the changes in my branch, I will open another PR.

PR #192, opened 2017-06-11
Add `naaative` theme

jorvi:
Naaative is based on `steeef` with a few minor changes:

* Adapted to look best with the 'Oceanic Next Dark' theme, but will look nice with almost any theme
* Use the Terminal theme colors instead of absolute (256) colors
* Displays the same Git indicators as a lot of custom prompts do (! for untracked changes, + for uncommitted changes, ? for untracked files)
* Adds a stash indicator ($)

Cheers :)

### DIFF
--- a/modules/prompt/functions/prompt_naaative_setup
+++ b/modules/prompt/functions/prompt_naaative_setup
@@ -1,0 +1,1 @@
+/Users/jvisser/projects/zim/modules/prompt/themes/naaative.zsh-theme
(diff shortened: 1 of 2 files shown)


ericbn:
I would refrain from accepting PRs for new themes (mainly customized versions of already existing ones) until we migrate to the zimframework organization.

Eriner:
... Which, if you'll be around this weekend @ericbn, I'd like to do.

alexweber:
Any updates on this?  I was considering adding a customized version of steeef myself but this might be just what I'm looking for...

ericbn:
I still think that the themes that come with zim, alongside our git-info module, provide a good basis for users to customize their own themes, and that we should not include variations of the already existing themes to our repo.

Eriner:
I agree with @ericbn, and would be happy to include this as an external module if @jorvi would like to maintain this in a separate repo.

Issue #184, opened 2017-05-21
Deprecate support for zsh versions < 5.2


ericbn:
As discussed in #177, and as per [last Eriner's comment](https://github.com/Eriner/zim/pull/177#issuecomment-302901655):

> I would suggest we deprecate support for < 5.3 in the init file now, and in two weeks on June 3rd, we will begin allowing/merging breaking changes for < 5.3.
>
> The majority of users should already be on a current version of zsh, but if they aren't, two weeks is plenty of time of advance notice. After that, we will return 1 on zsh != latest minor patch and merge breaking changes. Any issues with this?

lnicola:
One reason to support older versions of `zsh` would be the fact that `5.3` is pretty rare. CentOS 7 has `zsh 5.0` and `zim` already has issues with that (the `eriner` prompt no longer works, for example).

In the case of CentOS, there is a third-party repository with `zsh 5.1`, but I don't see myself upgrading to a newer version.

Eriner:
That means CentOS 7 matches approximate versions with Apple and Debian-stable - par for the course. Again, this is an upstream packaging issue. Some distros like CentOS and Debian-stable only maintain the most stable, vintage, and beautifully-aged packages to remove the burden of having to run modern software from their users. Maintaining full compatibility with these distros, working around code changes that were made over a year ago, is an necessary burden on developers. It makes sense in a server environment where stability is king, but who is installing zim on such a system? 

I'd be willing to create a branch of zim as it is now to maintain the old version, but no additional feature changes will be committed to support these old versions, and master will only support zsh >= 5.3.
We can prompt users to switch to this branch if zsh < 5.3.

If anyone wants to change nothing and have zim work with *-stable zsh packages, we'll have them pull from "legacy-5.0" branch.
For the rest (and target group) who want to have updated, current code, they'll pull master and either `make install` or install the relevant package to get a version of zsh that's been compiled within the last 365 days.

ericbn:
@Eriner, maybe we are restricting Zim to ArchLinux and MacOS users with this... (which we both are fine with, of course.)

And won't maintaining a separate "legacy-5.0" branch together with a master, for zsh latest version, be more work than just maintaining one branch? Also not sure this is desired for the future plans on the zimframework organization, as that would mean maintaining those branches for each separate module.

Eriner:
[...]

PR #194, opened 2017-06-28
Warning message for zsh < 5.2

ericbn:
as discussed in #184. We'll first give users this warning message. Next update will replace this by an error.

Also, starting at next update, a new branch will be created for users that still want to stick with zsh < 5.2.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -2,6 +2,11 @@
 # Zim initializition
 #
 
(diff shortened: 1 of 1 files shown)


ericbn:
@Eriner, maybe you'll want to polish the message to the users...

Eriner:
LGTM

PR #196, opened 2017-07-24
[git] Commit messages with 50 characters in glG

ericbn:
truncating at character 60. First 10 chars are used for `*`, commit hash and spaces.

The 50 char limit is based on http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Changed aliases to use the `--long-option` format when available, so aliases are clearer for those unfamiliar with the options.

Also changed `remote update` in gfu alias to `fetch --all`. Both behave the same way, fetching from all configured remotes. The `fetch --all` is shorter, and probably more familiar. See https://stackoverflow.com/a/17512004/2654518

### DIFF
--- a/modules/git/init.zsh
+++ b/modules/git/init.zsh
@@ -71,7 +71,7 @@ alias gf='git fetch'
 alias gfc='git clone'
 alias gfm='git pull'
 alias gfr='git pull --rebase'
-alias gfu='git remote update --prune; git merge --ff-only @\{u\}'
+alias gfu='git fetch --all --prune && git merge --ff-only @\{u\}'
 
 # Grep (g)
 alias gg='git grep'


ericbn:
Example with the change:

```
* f443b54 [git] Commit messages with 50 characters in glG    <Eric Nielsen> (89 seconds ago) (HEAD -> git-alias, ericbn/git-alias)
|
* d562acc [history-substring-search] Remove redundant defa.. <Eric Nielsen> (7 days ago) (origin/master, origin/HEAD, master)
|
* ff32968 Warning message for zsh < 5.2                      <Eric Nielsen> (4 weeks ago)
|
...
```

modules/git/init.zsh
```diff
@@ -71,7 +71,7 @@ alias gf='git fetch'
 alias gfc='git clone'
 alias gfm='git pull'
 alias gfr='git pull --rebase'
-alias gfu='git remote update -p; git merge --ff-only @\{u\}'
+alias gfu='git fetch --all --prune; git merge --ff-only @\{u\}'
```
ericbn: Can we change the `;` to `&&` so the merge only gets called if the fetch succeeded?

Eriner:
Yes, with that change, LGTM.

PR #199, opened 2017-07-30
Add brew completions

seanzxx:
Add brew completions. 
zsh-completions removed a lot of completions, e.g. docker, docker-machine.


ericbn:
I use zsh 5.3.1 installed by brew, and I already have `/usr/local/share/zsh/site-functions` in my `fpath`. Could not yet find where is that coming from...

Eriner:
https://unix.stackexchange.com/questions/26555/fpath-in-zsh-functions-and-site-functions#26558

PR #200, opened 2017-07-30
Don't load zim if zsh < 5.2

ericbn:
and show an error message, as discussed in #184.

Code in modules/environment/init.zsh is also simplified to only support zsh >= 5.2

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -4,7 +4,8 @@
 
 autoload -Uz is-at-least
 if ! is-at-least 5.2; then
-  print "WARNING: Support for zsh < 5.2 will be removed in a future update. Update your version of zsh now to prevent any interruptions." >&2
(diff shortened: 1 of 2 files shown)


init.zsh
```diff
@@ -3,49 +3,49 @@
...
-
-  for wanted_module (${zmodules}); do
-    if [[ -s "${ZIM}/modules/${wanted_module}/init.zsh" ]]; then
-      source "${ZIM}/modules/${wanted_module}/init.zsh"
-    elif [[ ! -d "${ZIM}/modules/${wanted_module}" ]]; then
-      print "No such module \"${wanted_module}\"." >&2
-    fi
-  done
-}
+if is-at-least 5.2; then
```
Eriner: rather than wrap this whole thing in a conditional, invert the test here and exit early.

ericbn:
@Eriner, updated the commit.

Eriner:
LGTM

PR #203, opened 2017-08-10
Allow arbitrary location by setting ZIM_HOME

m45t3r:
This change allow arbitrary ZIM location path by setting a new environment variable, ZIM_PATH. If the user does not set it, ZIM falls back to the old `"${ZDOTDIR:-${HOME}}/.zim"` location.

PR #210, opened 2017-09-04
Make zim location configurable and install to XDG_DATA_HOME by default

chetgurevitch:
Supersedes https://github.com/Eriner/zim/pull/203. Zims location can be set with ZIM_DATA_DIR and the default is XDG_DATA_HOME (~/.local/share) per the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

My only concern left with this PR is whether we want to print a message when zim is launched from the old directory. Users would have to update the location definition in their .zshrc but after that they could just run `mv ~/.zim ~/.local/share/zim`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,13 @@ Installing Zim is easy. If you have a different shell framework installed (like 
 
 1. In a Zsh shell, clone the repository:
   ```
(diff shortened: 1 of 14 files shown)


ericbn:
Closing this in favor of #203

PR #211, opened 2017-09-04
Use ZIM_DATA_DIR and install to XDG_DATA_HOME by default

chetgurevitch:
Zims location can be set with ZIM_DATA_DIR and the default is XDG_DATA_HOME (~/.local/share) per the XDG Base Directory Specification. 

I'd prefer this convention for the directory variable as in the xdg spec _HOME specifies only the default directory in a grouping of directories to be sourced simultaneously and ZIM_HOME doesn't clarify DATA vs CONFIG vs RUNTIME etc.

In addition after reading through the spec I'd say the zim folder itself definitely belongs in XDG_DATA_HOME rather than XDG_CONFIG_HOME. If anything were to go in the latter it would probably be .zimrc as ~/.config/zim/init.zsh, but I think it makes sense to leave .zimrc where and as it is unless upstream zsh changes its dotfile configuration. This PR also includes all necessary readme and template changes.

I'd greatly appreciate input on the above.

A concern left with this PR is whether we want to print a message when zim is launched from the old directory. Users would have to update the location definition in their .zshrc but after that they could just run mv ~/.zim ~/.local/share/zim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,13 @@ Installing Zim is easy. If you have a different shell framework installed (like 
 
 1. In a Zsh shell, clone the repository:
   ```
-  git clone --recursive https://github.com/Eriner/zim.git ${ZDOTDIR:-${HOME}}/.zim
+  git clone --recursive https://github.com/Eriner/zim.git "${XDG_DATA_HOME:-"${HOME}/.local/share"}/zim"
   ```
 
 2. Paste this into your terminal to prepend the initialization templates to your configs:
   ```
   setopt EXTENDED_GLOB
-  for template_file ( ${ZDOTDIR:-${HOME}}/.zim/templates/* ); do
+  for template_file ( ${ZIM_DATA_DIR}/templates/* ); do
     user_file="${ZDOTDIR:-${HOME}}/.${template_file:t}"
     touch ${user_file}
     ( print -rn "$(<${template_file})$(<${user_file})" >! ${user_file} ) 2>/dev/null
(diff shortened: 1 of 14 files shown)


ericbn:
I think the XDG Base Directory Specification is a nice to have in the project, but I'm not in favor of emerging it all the way to the surface. I believe that users that don't want to use it, or don't even know how to use it, should not be forced to be exposed to the details.

For example, you're changing the `README.md` with specific information, and changing the environment variable to `ZIM_DATA_DIR`, which is also XDG-specific.

chetgurevitch:
Tbh I don't really understand your sentiment - the readme adjustment is only a tweak to the default installation location and existing installations continue to function just fine. Also on the variable naming I'd say it seems pretty in line with ZDOTDIR for example but with the underscores as the project already seemed to be doing. I could certainly remove the name chanage from the pr and I suppose the XDG config check from the install instructions...

Eriner:
FWIW, as I have mentioned previously, I am all for supporting the XDG spec. However, I also think that we shouldn't force this on to users until upstream zsh does the same. In otherwords, until zsh requires that user configuration be defined in accordance to the XDG spec, we shouldn't require this either.

Litmus test: once `.zshrc` is in `${HOME}/.config/zsh/zshrc`, we'll enforce this in the project.

Eriner:
Also, as a critique of the code itself, until we can enforce the location change everywhere, we'd be required to (as you've done) set a variable and check for the location of the files in a conditional operation on every init, which seems unnecessary to me.

Eriner:
I appreciate the PR and code contribution, and perhaps one day when upstream ZSH decides to conform to the XDG spec, we'll be able to merge this. However, until that point, I'll be holding off on "total conversion" code such as this.

The `${ZDOTDIR}` that you've replaced is defined [upstream](http://zsh.sourceforge.net/Intro/intro_3.html). Once this is changed and zsh conforms to the XDG spec, I'd be happy to re-open this PR.

Please don't take this personally. I'd love for every utility I use to conform to XDG specs so that I always knew where to go to find the config and data files, but alas, adoption (and time) are the limiting factors here. This is something I'd like to move on in the future, when upstream zsh makes a move.

PR #215, opened 2017-09-25
[template] Add zcompile of git-info to zlogin

ericbn:
and declare `zim_mods` as a local variable. Also minor formatting changes.


Eriner:
:+1:

Changes like this are good, especially now that we're not delegate to supporting the legacy `local` declaration syntax issues that arise when using <5.2.

LGTM

PR #216, opened 2017-09-26
[template] Use zrecompile and create digest zwc file

### LINKED ISSUE #86: Investigate usage of zrecompile for zlogin
See documentation here: http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Recompiling-Functions

Currently, `.zlogin` checks specific functions that are "heavy" and could actually benefit from `zcompile`. I have also implemented an "if source newer than compiled" function to check if zcompile is required. Seemingly, this functionality is already provided by zsh and need only be autoloaded. My current implementation can be seen below:

```
  # Function to determine the need of a zcompile. If the .zwc file
  # does not exist, or the base file is newer, we need to compile.
  # These jobs are asynchronous, and will not impact the interactive shell
  zcompare() {
    if [[ -s ${1} && ( ! -s ${1}.zwc || ${1} -nt ${1}.zwc) ]]; then
      zcompile ${1}
    fi
  }
```

Replacing `zcompare` with `zrecompile` (or a loose wrapper) may or may not be useful. This is something to be investigated.

ericbn:
for the autoloaded functions. According to the `zshbuiltins` documentation:

    Files containing multiple compiled functions are called `digest'
    files,  and are intended to be used as elements of the FPATH/fpath
    special array.

Also compiling all module init files, instead of some selected ones. This didn't impact performance much (time to run async block in zlogin when from 0.22s to 0.25s on my machine).

Didn't manage to make adding the prompt `prompt_*_setup` functions to the digest file work, so kept these separate.

Closes #86

### DIFF
--- a/templates/zlogin
+++ b/templates/zlogin
@@ -8,52 +8,38 @@
 #
 
 (
-  # Function to determine the need of a zcompile. If the .zwc file
-  # does not exist, or the base file is newer, we need to compile.
-  # These jobs are asynchronous, and will not impact the interactive shell
-  zcompare() {
-    if [[ -s ${1} && ( ! -s ${1}.zwc || ${1} -nt ${1}.zwc ) ]]; then
-      zcompile ${1}
-    fi
-  }
-
   local zim_mods=${ZIM_HOME}/modules
   setopt EXTENDED_GLOB
+  autoload -U zrecompile
 
   # zcompile the completion cache; siginificant speedup.
   for file in ${ZDOTDIR:-${HOME}}/.zcomp^(*.zwc)(.); do
-    zcompare ${file}
+    zrecompile -pq ${file}
   done
 
   # zcompile .zshrc
-  zcompare ${ZDOTDIR:-${HOME}}/.zshrc
+  zrecompile -pq ${ZDOTDIR:-${HOME}}/.zshrc
 
-  # zcompile some light module init scripts
-  zcompare ${zim_mods}/git/init.zsh
-  zcompare ${zim_mods}/git-info/init.zsh
-  zcompare ${zim_mods}/utility/init.zsh
-  zcompare ${zim_mods}/pacman/init.zsh
-  zcompare ${zim_mods}/spectrum/init.zsh
-  zcompare ${zim_mods}/completion/init.zsh
-  zcompare ${zim_mods}/fasd/init.zsh
-
-  # zcompile all .zsh files in the custom module
-  for file in ${zim_mods}/custom/**/^(README.md|*.zwc)(.); do
-    zcompare ${file}
+  # zcompile all module init scripts
+  for file in ${zim_mods}/*/init.zsh; do
+    zrecompile -pq ${file}
   done
 
   # zcompile all autoloaded functions
-  for file in ${zim_mods}/**/functions/^(*.zwc)(.); do
-    zcompare ${file}
+  zrecompile -pq ${ZIM_HOME}/functions ${zim_mods}/*/functions/^([_.]*|prompt_*_setup|README*|*.zwc)(.)
+
+  # zcompile all prompt setup scripts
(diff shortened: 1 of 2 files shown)


templates/zlogin
```diff
@@ -8,52 +8,38 @@
...
-
   local zim_mods=${ZIM_HOME}/modules
   setopt EXTENDED_GLOB
+  autoload -U zrecompile
 
   # zcompile the completion cache; siginificant speedup.
   for file in ${ZDOTDIR:-${HOME}}/.zcomp^(*.zwc)(.); do
-    zcompare ${file}
+    zrecompile -pq ${file}
   done
```
ericbn: I see the `.zcompdump` file is already being compiled in `modules/completion/init.zsh`, and it's considering the `zcompdump_file` configuration variable there, which is a more complete solution

templates/zlogin
```diff
@@ -8,52 +8,38 @@
...
+  # zcompile all module init scripts
+  for file in ${zim_mods}/*/init.zsh; do
+    zrecompile -pq ${file}
   done
 
   # zcompile all autoloaded functions
-  for file in ${zim_mods}/**/functions/^(*.zwc)(.); do
-    zcompare ${file}
+  zrecompile -pq ${ZIM_HOME}/functions ${zim_mods}/*/functions/^([_.]*|prompt_*_setup|README*|*.zwc)(.)
+
```
ericbn: Now I see an issue with this approach: it will make available all module functions, even for the modules that are not loaded.

Eriner: perhaps wrap this in a for-loop that iterates over `zmodules` instead of wildcarding the path after `${zim_mods}`?

Eriner: the other downside to this, is when changes are made to the config file, we'll have to `source .zlogin` to regen everything; they will no longer be a part of fpath (should probably make a zmanage convenience function for this; perhaps "rebuild" or "compile"?)

templates/zlogin
ericbn: Can I remove this block? See #217

templates/zlogin
ericbn: Only including `zmodules` fixes this. And `zrecompile` is smart enough to recompile the digest when the list of files changes. 

I'd like to apply the same approach to the init files, and only compile those for the enabled `zmodules`.

I don't get the downside you mentioned. In my understanding, in the current implementation a `source ~/.zlogin` is already needed to load new modules, and only starting a new shell would work to "not load" any removed modules. In the new solution, `fpath` never changes, not even when `zmodule` changes, so maybe this is even a upside...  :- )

Eriner: So right now, if we compile all of the modules into a single digest and add that digest to the fpath, we've enabled all of the modules. The `zmodules` content isn't considered; they're all enabled.

To be clear, I think we should only compile functions and init scripts for modules enabled in the `zmodules` array.

ericbn:
Added a new commit with these changes:

* zcompile only enabled modules and don't zcompile the `.zcompdump` file in zlogin, since it's already being compiled by the completion module. We've been trying to zcompile it twice before this change.

* Also refactor the zim `init.zsh` script, and do not unset `zmodules` anymore because we need it in zlogin.

ericbn:
Maybe we should move the `zlogin` logic to a `${ZIM_HOME}/login_init.zsh` script, and make the `zlogin` template to only contain

    [[ -s ${ZIM_HOME}/login_init.zsh ]] && source ${ZIM_HOME}/login_init.zsh

so `login_init.zsh` is under version control and does not need to be manually updated by the user, as now is the case with `~/.zlogin` and `${ZIM_HOME}/templates/zlogin`.

Eriner:
@ericbn that's a really good idea.

ericbn:
Done in #218. Closing this PR in favor of the new one.

PR #217, opened 2017-09-26
[completion] Consider the zcompdump_file variable

ericbn:
that could be defined in `.zimrc`, when doing the zcompile in `modules/completion/init.zsh` and cleaning the cache.

A custom `${zcompdump_file}` would not be zcompiled or cleaned from the cache without this correction.

Also, use zrecompile instead of the custom logic to call zcompile.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -23,13 +23,8 @@ source ${0:h}/compdefs.zsh
 
 {
   # zcomple the .zcompdump in the background
-  zcompdump=${ZDOTDIR:-${HOME}}/.zcompdump
-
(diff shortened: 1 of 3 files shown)


Eriner:
I'll spend some time looking at this once we've okay'd #216, that way we can then merge these together (as they both use `zrecompile`.

ericbn:
Maybe we should move this zcompile of the zcompdump_file to the `zlogin` script, so all compilation happens in one place.

Eriner:
The only issue with that, and hence why it's currently not part of the `.zlogin` compilation, is that if you `zcompile` it before the shell has initialized, you'll miss functions that are dynamically created by other plugins (like the history completion funcs).

Eriner:
This may not be _that_ big of a deal, and the volume of additional functions may be few enough that it makes more sense to eat that just for the sake of code simplification.

ericbn:
Cool. Closing this PR is favor of #218.

PR #219, opened 2017-09-29
Add Control+A and Control+E shortcuts

hajzso:
Add Control+A and Control+E shortcuts for beginning of line and end of line.

I've been using these forever, and they are standard in a lot of terminals, I think it would be nice to set them in this module for people who are used to it.


Eriner:
Just add `bindkey -e` after you `source` zim in your `.zshrc`.

hajzso:
Didn't know about this one, nice! thanks :+1:

PR #221, opened 2017-10-08
Support for multipart 7z archives

djape:
Add support for unarchiving of multipart 7z archives.
Example:
`unarchive multipart.7z.001`

### DIFF
--- a/modules/archive/functions/unarchive
+++ b/modules/archive/functions/unarchive
@@ -37,6 +37,6 @@ case "${archive_name}" in
   (*.rar) unrar &> /dev/null \
     && unrar x -ad "${archive_name}" \
     || rar x -ad "${archive_name}" ;;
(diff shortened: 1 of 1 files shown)


ericbn:
LGTM. Could be done for `*.zip.001` too...

Eriner:
and rar...

djape:
@ericbn AFAIK`zip -s` creates `multipart.z0x, ...` and `multipart.zip`, but `unzip` does not support multipart archives yet (you have to concatenate files first). `7z` can split archives using different formats, so maybe `(*.7z|*.001) 7za x "${archive_name}"` should be used. What do you think?

ericbn:
> so maybe `(*.7z|*.001) 7za x "${archive_name}"` should be used. What do you think?

Good point, @djape: the `*.001` archives are specific to 7zip. I agree you suggestion.

djape:
Done.

PR #224, opened 2017-11-01
[spectrum] Delete submodule

ericbn:
The following was introduced in zsh 5.0:

> %F, %K, %f, %k allow colouring of prompts both using ANSI colour names and values supported by individual terminal emulators.

making this submodule obsolete.

See http://zsh.sourceforge.net/releases.html


Eriner:
:+1: LGTM!

PR #225, opened 2017-11-20
[utility] BSD uses LSCOLORS and GREP_COLOR

ericbn:
and GNU uses LS_COLORS and GREP_COLORS. So change logic to only set these environment variables for each OSTYPE.

Also original code was setting the `ls` alias with `-G` instead of `--color=auto` if in a GNU without `dircolors`. Fix that too.

Simplify arithmetic command expression without using `${var}`, but `var` instead. The second construct does not fail if `var` is not defined, but defaults its value to `0` in that case.

PR #228, opened 2017-11-24
Change color red for decorated text to yellow as it different when ru…

well1791:
Hi 👋 

This contribution might not bee too much but I think it's just a visual enhancement.

As you know, when running `gld` the color for old changes is red which is also the same color for `--decorated` option.

Now, as I changed it to yellow the difference is more notable. Also, for dark themes/backgrounds that red color is a bit hard to read as it's not contrasting with a lighter color.

Here the differences: 

image

Please let me know your thoughts about this changes.
Also, thanks for this amazing repo! 👏


ericbn:
I'm using Solarized Dark and I'm lightly color-blind, so it's a subtle difference to me. But I like it! LGTM

ericbn:
I just gave this a second look, and there's a more interesting option for when coloring the ref names (`%d`): it is `%C(auto)`. This is already being used in the `_git_log_fullgraph_format`

Then I checked what does `%C(auto)` renders for the other elements we're using, and I just got an yellow color for the commit hash (`%h` or `%H`). Everything else was just my default foreground color (besides the ref names, of course).

So I though, "oh nice thing we have more colors than 'auto' gives us"...

Then I saw we're coloring the author as cyan and the date as blue in `_git_log_medium_format`, and on the other hand coloring the author as bold blue and the date as yellow in `_git_log_fullgraph_format`.

And I though "what a mess"...  :- )

So trying to unify the colors I'm suggesting:
* yellow for the commit hash (because that's what auto gives me already for this)
* green for the author (I think green gives more emphasis than blue or cyan that we used before, and less than yellow)
* blue for the date (less emphasis, and also we already use it in `_git_log_medium_format`)
* auto for the ref names (`%d`)

I ended up with this:
```
# Log
_git_log_medium_format='%C(bold)Commit:%C(reset) %C(yellow)%H%C(auto)%d%n%C(bold)Author:%C(reset) %C(green)%an <%ae>%n%C(bold)Date:%C(reset)   %C(blue)%ai (%ar)%C(reset)%n%+B'
_git_log_oneline_format='%C(yellow)%h%C(reset) %s%C(auto)%d%C(reset)%n'
_git_log_fullgraph_format='%C(yellow)%h%C(reset) %<|(60,trunc)%s %C(green)<%an> %C(blue)(%cd)%C(auto)%d%C(reset)%n'
_git_log_brief_format='%C(yellow)%h%C(reset) %s%n%C(blue)(%ar by %an)%C(auto)%d%C(reset)%n'
```

See [docs](https://git-scm.com/docs/pretty-formats) for reference.

@well1791, what do you think? (Since I see you're good with details too)  :- )
@Eriner?

well1791:
Hi @ericbn !

Actually I didn't noticed those details you mentioned, but thanks for being consistent with the code.

Also, I tested the colors you suggested and after working a couple of hours with it!
I must say you have good eyes for colors and `auto` looks good btw.

As I hope everyone agree on this settings, let me know if it looks good to you and @Eriner so I can add this changes to the PR (or just feel free to commit changes).

Cheers!

ericbn:
@well1791, I never payed attention to those either before...

Also, thanks to @hypnoglow for creating the `glG` alias that had `%C(auto)` on the first place (at 3c0e49a).

Here are different variations in Solarized Light and Solarized Dark for comparison:

### Original

screen shot 2017-11-25 at 10 21 03

### Suggestion 1

Yellow commit hash, green author, blue date, auto for the ref names

screen shot 2017-11-25 at 10 17 28

Suggestion 1 is closer to what we have in `glG`.

### Suggestion 2

Like above, but cyan author

screen shot 2017-11-25 at 10 15 35

Suggestion 2 is closer to what we have in `gl`.

well1791:
I vote for suggestion 2 :+1: blue is my fav color and cyan is close to it :blue_heart:

Eriner:
Colours look fine to me -- I don't really have a preference either way.

modules/git/init.zsh
```diff
@@ -7,10 +7,10 @@
 #
 
 # Log
-_git_log_medium_format='%C(bold)Commit:%C(reset) %C(green)%H%C(red)%d%n%C(bold)Author:%C(reset) %C(cyan)%an <%ae>%n%C(bold)Date:%C(reset)   %C(blue)%ai (%ar)%C(reset)%n%+B'
-_git_log_oneline_format='%C(green)%h%C(reset) %s%C(red)%d%C(reset)%n'
-_git_log_fullgraph_format='%C(green)%h%C(reset) %<|(60,trunc)%s %C(bold blue)<%an>%C(reset) %C(yellow)(%cd)%C(reset)%C(auto)%d%C(reset)%n'
-_git_log_brief_format='%C(green)%h%C(reset) %s%n%C(blue)(%ar by %an)%C(red)%d%C(reset)%n'
+_git_log_medium_format='%C(bold)Commit:%C(reset) %C(yellow)%H%C(auto)%d%n%C(bold)Author:%C(reset) %C(cyan)%an <%ae>%n%C(bold)Date:%C(reset)   %C(blue)%ai (%ar)%C(reset)%n%+B'
```
ericbn: 👍 @well1791, can we have a comment above here saying

    # Colour scheme has yellow commit hash, cyan author, blue date, auto ref names

?

ericbn:
@lnicola, since you have a terminal with a dark blue, I'm curious on how does the `gl`, `glb` and `glG` commands look like (when executed inside a git repo), after the change done here.

lnicola:
Right, the blue isn't too pretty:

image
image
image

ericbn:
@lnicola, have you considered changing your terminal colors? Why do you like that blue if you cannot see it? 😄

lnicola:
No, because I never saw it until updating `zim`. Oh wait 😄.

It's only been an issue before in `vim` until I found out about `set background=dark`.

ericbn:
Yeah, I see `vim`'s default colors in that case have all blues as bold. Interesting. Your terminal only has 15 colors. 🤔

lnicola:
I'd change it, but I don't know what color to pick 😃, and I'd rather not lose too much time with this. Other terminal themes look better.

But you had the same issue here: https://github.com/Eriner/zim/pull/225#issuecomment-346210796.

ericbn:
@lnicola, fair enough, I'm in favor of replacing "blue" to "bold blue" here, and mainly at #225.

CC @well1791

Thoughts?

well1791:
Make sense.

Actually, I was having a bit of trouble there too when using Linux Mint, so I changed the palette color.

(I'll paste an image later)

But, when working with Mac (iTerm2), it deals with colors pretty good. No need to change anything.

image

Regarding vim color, I use [ir_black from spf13](https://github.com/spf13/vim-colors/blob/master/colors/ir_black.vim)

image

Anyway, I agreed with bolding the Author name 👍 (sorry for the long post)

ericbn:
Just read a nicely written article that reminded me about the discussion on the blue color we had in this issue. Here's the article: https://bsago.me/posts/that-annoying-shade-of-blue

lnicola:
Thanks, I hate it.

CC @romkatv your prompt theme has a pretty awful color scheme (black on red background or something like that) when `$PWD` is `/etc`.

romkatv:
> Thanks, I hate it.
> 
> CC @romkatv your prompt theme has a pretty awful color scheme (black on red background or something like that) when `$PWD` is `/etc`.

Do you mean powerlevel10k with default settings, powerlevel10k with preconfigured styles or perhaps powerlevel10k with my personal config? Neither seems to show anything in red when the current directory is `/etc`.

Default:

image

Lean:

image

Classic:

image

Rainbow:

image

Pure:

image

My personal config (Lean with some options from `p10k configure`):

image

That said, nothing is stopping anyone from configuring powerlevel10k the way you described.

lnicola:
I meant the default settings. I was using an older version, from before the configuration wizard. I can no longer find that specific instance, although it's still using the blue from this discussion:

image

Anyway, I'll try to set up p10k again and report on your issue tracker if I find another egregious color combination, since it's completely off-topic here.

romkatv:
I'll try to sum up, let me know if I got it right.

1. Powerlevel10k does not display directory in red when the current directory is `/etc` unless users explicitly configure it this this.
2. Some terminals have unpleasant base colors (0 through 7) and/or some users configure their terminals this way.

romkatv:
@lnicola I think there is some confusion. The colors in Rainbow and Pure styles follow the color pallet of your terminal. Colors in Lean and Classic styles are fixed. (If you select Lean, there is a choice between fixed colors and terminal-pallet-respecting colors; the first screen shows the fixed variant). If your terminal comes preconfigured with an awful color scheme (very few terminals do), it's a good idea to change it.

lnicola:
Yeah, the colors are better since you fixed https://github.com/romkatv/powerlevel10k/issues/426. I think it wasn't fully loading its settings, since my prompt is also different now (no username and hostname).

I _should_ change my terminal colors, but I'm not sure what to pick :confused:.

romkatv:
> Yeah, the colors are better since you fixed [romkatv/powerlevel10k#426](https://github.com/romkatv/powerlevel10k/issues/426). I think it wasn't fully loading its settings, since my prompt is also different now (no username and hostname).
> 
> I _should_ change my terminal colors, but I'm not sure what to pick .

The most popular color schemes are Solarized and Tango. Snazzy is decent. There is plethora of choices if you google "terminal color scheme".

lnicola:
> The most popular color schemes are Solarized and Tango. Snazzy is decent. There is plethora of choices if you google "terminal color scheme".

I think I'm using Solarized dark from Gnome Terminal.

romkatv:
Solarized Dark in GNOME Terminal looks like this:

image

From the looks of it, you are using only the base colors (the color of terminal background and unstyled text) from Solarized Dark, and your color palette is "Linux console".

PR #229, opened 2017-11-27
[completion] Remove smart editor completion

ericbn:
as Vim can open archives, and Emacs can virtually open everything  :- )

I suggest we don't try to be smart here.


Eriner:
This is something that I personally like (I'll only ever open code/text/no-extension files in my editor), especially in "busy" directories. This is a change that I can move to my personal configs and I can see why this might result in confusion on behalf of some users. LGTM.

PR #237, opened 2017-12-22
Improve issue template

deltaxflux:
Hey,
I change the issue template. It is now easier for a developer to read because only the user can read the HTML comments. 

All content inside  `<!-- -->` is only visible to users. So you have less unnecessary information.

Regards, 
_deltax_

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,26 +1,38 @@
+<!-- 
+Note:
+
 Please check the existing issues to make sure you're not duplicating a report.
 
 For bug reports, please provide the following information:
 
-In a terminal, run `zmanage info` and paste the output below:
+In a terminal, run `zmanage info` and paste the output below: -->
 
+#### Details:
 ```
 #Paste the output here
 ```
 
-- gist of .zimrc:
-- gist of .zshrc:
+##### Gist of .zimrc:
+  ```
+  #Paste the output here
+  ```
+##### Gist of .zshrc:
+  ```
+  #Paste the output here
+  ```
 
-Description
------------
+#### Description
 
+#### Do you want to request a *feature* or report a *bug*?
+*Bug*
 
-Steps to reproduce
-------------------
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,26 +1,38 @@
...
+Note:
+
 Please check the existing issues to make sure you're not duplicating a report.
 
 For bug reports, please provide the following information:
 
-In a terminal, run `zmanage info` and paste the output below:
+In a terminal, run `zmanage info` and paste the output below: -->
 
+#### Details:
```
ericbn: Maybe the sentence "In a terminal, ... and paste the output below:" could be changed to a comment inside a header like

    #### Output of `zmanage info`:
      ```
      #In a terminal, run `zmanage info` and paste the output here
      ```

And then there would be no `#### Details` header.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,26 +1,38 @@
...
+  ```
+  #Paste the output here
+  ```
+
+#### Description
 
-Description
------------
+#### Do you want to request a *feature* or report a *bug*?
+*Bug*
```
ericbn: Maybe we don't need this, since the two following headers are mutually exclusive in this sense.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,26 +1,38 @@
...
 
+#### If the current behaviour is a bug, please provide the steps to reproduce it.
+<!-- A great way to do this is to provide screenshots and commands.* -->
```
ericbn: Extra `*` at the end of the sentence?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,26 +1,38 @@
...
 
-Steps to reproduce
-------------------
+#### If this is a feature request, what is the motivation or use case for the added feature?
 
+#### Images or other information
```
ericbn: Not sure if we need this...

@Eriner, thoughts?

PR #244, opened 2018-01-11
[git] Breaking changes in aliases and functions

ericbn:
Rename `gpS` (mnemonic "pretty-signature") and `glS` ("log-signature") to `gcv` ("commit-verify") and `glv` ("log-verify"), so `v` is used to display/validate signatures, and `s`/`S` is used to sign.

Remove `git-commit-lost` and `git-stash-dropped`, as this could be probably joined into one function that shows all lost commits, including lost stashes, and also because there seems to be many different ways to do this, as discussed in https://stackoverflow.com/q/89332/2654518.  An answer worth noting is https://stackoverflow.com/a/6455586/2654518.  If someone is using this, let's discuss on a new solution...

Remove Oh-My-Zsh compatibility from `git-branch-current`, and delete `git_current_branch`. This compatibility was already partially broken at c1b02c0, when we also updated the gitster prompt theme. Now this function fails when in 'detached HEAD' state (is was returning the commit hash in this state), which is what we would expect from the use we make of it.

**EDIT**: Added more changes:

Remove `glb` alias and add `glO`, as there's actually no "brief" format originally in git (by "originally" I mean the ones listed for `--format=<format>` in the `git log` help). Also, we already fit all the information that `glb` had in the `glG`. So add `glO`, that is a one-line variant of `glG` (which displays in graph format). Now we get nice symmetrical formats in `glg`/`glG` and `glo`/`glO`.


ericbn:
@Eriner, I would like your thoughts here, as I'm proposing changes to some aliases you created. 

And I noticed there's the `git-ignore-add` function that is not being referenced in any alias. Should we have an alias for it? Should we remove it?

(Also, I did some [non-breaking] changes at 0a74519 and c501a1b that you might want to comment on too)

modules/git/init.zsh
```diff
@@ -45,9 +45,8 @@ alias gcP='git cherry-pick --no-commit'
 alias gcr='git revert'
 alias gcR='git reset "HEAD^"'
 alias gcs='git show --pretty=format:"${_git_log_medium_format}"'
-alias gcl='git-commit-lost'
 alias gcS='git commit -S'
-alias gpS='git show --pretty=short --show-signature'
+alias gcv='git verify-commit'
```
ericbn: I've changed from `git show --show-signature` to `git verify-commit` because we're already using `git verify-tag` at `gtv`.

Eriner:
@ericbn LGTM

PR #245, opened 2018-01-12
[prompt] steef: Update it to be customizable

ericbn:
and to have a help and a preview too.

Inspired by #192, and by having (few but) good prompt themes to serve as
starting points for the users to customize or implement their own.


ericbn:
@jorvi and @alexweber, since you were interested in customizing the steeef prompt theme, and I'm trying to make our themes customizable using the `promptinit` zsh function, I'd like your thought on these changes and if they would serve your customization needs...

alexweber:
@ericbn This is great, thanks!!

PR #246, opened 2018-02-12
Repair gCl alias

sQVe:
I noted that I got a `invalid option` error when trying to run the alias `gCl`. The issue with this alias is that the `no-pager` option was set on `git-diff` instead of `git`.


ericbn:
Thanks for the fix, @sQVe.

Issue #135, opened 2017-01-26
gpg-agent


beckend:
Is there a gpg-agent plugin?

Eriner:
Currently, no.

Tuurlijk:
@beckend what should the plugin provide?

beckend:
Something like:
```bash
# Add the following to your shell init to set up gpg-agent automatically for every shell
if [ -f ~/.gnupg/.gpg-agent-info ] && [ -n "$(pgrep gpg-agent)" ]; then
    source ~/.gnupg/.gpg-agent-info
    export GPG_AGENT_INFO
else
    eval $(gpg-agent --daemon --write-env-file ~/.gnupg/.gpg-agent-info)
fi
export SSH_AUTH_SOCK=$HOME/.gnupg/S.gpg-agent.ssh
```

Tuurlijk:
I see. We'll need to adjust this a bit to also make it work for gpg v 2. This gives me: gpg-agent[xxx]: WARNING: "--write-env-file" is an obsolete option - it has no effect

Think we should use gpgconf in that case

AtomicCoding:
Why not just use the [`oh-my-zsh` gpg-agent plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/gpg-agent) or [this](https://github.com/axtl/gpg-agent.zsh)?

Brottweiler:
You could probably use a plugin manager and do something like this `zplug "plugins/gpg-agent",   from:oh-my-zsh`, at least if you're using https://github.com/zplug/zplug

AtomicCoding:
@ericbn @Eriner, can we close this now that #257 is merged?

PR #250, opened 2018-03-22
Updated Plugin Loading

AtomicCoding:
Updated the plugin loading system to load plugins with `{plugin name}.plugin.zsh` and `{plugin name}.zsh` in order to improve compatibility.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+Please check the existing issues to make sure you're not duplicating a report.
+
+For bug reports, please provide the following information:
+
+In a terminal, run `zmanage info` and paste the output below:
+
+```
+#Paste the output here
+```
+
+- gist of .zimrc:
+- gist of .zshrc:
+
+Description
+-----------
+
+
+Steps to reproduce
+------------------
+
+
+Images or other information
+---------------------------
+
+
+
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,1 @@
+Please follow Zim's [code style guidelines](https://github.com/zimfw/zimfw/wiki/Code-Style-Guide).
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+*.swp
+*.zwc
+*.zwc.old
+custom/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,21 @@
+[submodule "modules/history-substring-search/external"]
+	path = modules/history-substring-search/external
+	url = https://github.com/zsh-users/zsh-history-substring-search.git
+[submodule "modules/completion/external"]
+	path = modules/completion/external
+	url = https://github.com/zsh-users/zsh-completions.git
+[submodule "modules/syntax-highlighting/external"]
+	path = modules/syntax-highlighting/external
+	url = https://github.com/zsh-users/zsh-syntax-highlighting.git
+[submodule "modules/prompt/external-themes/pure"]
+	path = modules/prompt/external-themes/pure
+	url = https://github.com/sindresorhus/pure.git
+[submodule "modules/prompt/external-themes/liquidprompt"]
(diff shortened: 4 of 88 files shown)


AtomicCoding:
@Brottweiler Sorry! Working on it right now!

AtomicCoding:
Fixed it! Sorry!

AtomicCoding:
@ericbn @Eriner Ping

ericbn:
I would not merge the logo. I like the idea for the init code. Actually I was working on something similar:

    for zmodule_file (${zmodule_dir}/${zmodule}.zsh-theme ${zmodule_dir}/init.zsh ${zmodule_dir}/${zmodule}.zsh ${zmodule_dir}/${zmodule}.sh ${zmodule_dir}/${zmodule}.plugin.zsh); do
      [[ -f ${zmodule_file} ]] && source ${zmodule_file} && break
    done

@Eriner, thoughts?

AtomicCoding:
@ericbn Just made some changes to prioritize `.plugin.zsh`. And prepend zsh- to the plugin script if all else fails.

Eriner:
From a performance perspective, due to the structuring of the conditionals, this won't cause any unecessary disk I/O. I haven't read over the code maticuliously and I'll wait to see what @ericbn says, but the concept and basic structure has a :+1: from me.

AtomicCoding:
@ericbn @Eriner 
Would it be ok to replace the last `elif` with an `else`, I think it would be better to tell people if a plugin isn't being loaded rather than if a plugin's directory doesn't exist.

Eriner:
Closes #135

ericbn:
@Eriner, I'm more in favor of the `for` notation, as we can take advantage of globs and it also looks more concise. What do you think?

I like the idea of trying to match a `zsh-` prefix too. This is what I suggest:

    local zmodule_dir
    # ...
    zmodule_dir=${ZIMFW_HOME}/modules/${zmodule}
    if [[ ! -d ${zmodule_dir} ]]; then
      print "No such module \"${zmodule}\"." >&2
    else
      for zmodule_file (${zmodule_dir}/{,zsh-}${zmodule}.zsh-theme \
          ${zmodule_dir}/init.zsh \
          ${zmodule_dir}/{,zsh-}${zmodule}.{z,}sh \
          ${zmodule_dir}/{,zsh-}${zmodule}.plugin.zsh); do
        if [[ -f ${zmodule_file} ]]; then
          source ${zmodule_file}
          break
        fi
      done
    fi

@AtomicCoding, it is valid to have for example a module with just a `functions` subdir, but no init script, so that's why we don't fail if that case.

---

@Eriner, a side note partially related to this change:  I've been actually working on the "distributed modules" loading mechanism for zim, and I extracted the code above from that. Maybe I should push that to a `develop` branch so we can start looking at it together...

AtomicCoding:
@ericbn I changed some of the priorities and had zmodule_dir be immediately set after being declared because (this might just be me) ZSH echoing the value of it for some reason when it wasn't set like that, either way, here it is:
```zsh
    local zmodule_dir=${ZIM_HOME}/modules/${zmodule}
    if [[ ! -d ${zmodule_dir} ]]; then
      print "No such module \"${zmodule}\"." >&2
    else
      for zmodule_file (${zmodule_dir}/init.zsh \
          ${zmodule_dir}/{,zsh-}${zmodule}.plugin.zsh \
          ${zmodule_dir}/{,zsh-}${zmodule}.zsh-theme \
          ${zmodule_dir}/{,zsh-}${zmodule}.{z,}sh); do
        if [[ -f ${zmodule_file} ]]; then
          source ${zmodule_file}
          break
        fi
      done
    fi
```

AtomicCoding:
@ericbn Most plugins usually have an init.zsh or (module name).plugin.zsh, so why not prioritize those?

AtomicCoding:
@ericbn @Eriner Just committed the for loop system. I'm using it on my installation and all seems to be ready to merge!

AtomicCoding:
Sorry, messed up the pull request

AtomicCoding:
Reopening it soon...

AtomicCoding:
New pull request at #257. Sorry, tried to clear commit tree. Can be merged cleanly now, however.

PR #252, opened 2018-03-29
hard-fail on missing zshbuiltin is-at-least

### LINKED ISSUE #249: You're using zsh version 5.4.1, and versions < 5.2 are not supported. Update your zsh.
Hi,
I'm running zim in Ubuntu 16.04. Everything was fine until I run "sudo apt autoremove" to clear non needed packages.
Now, when I open a terimal I get:
```
/etc/zsh/zshrc:111: compinit: function definition file not found
/home/cristian/.zim/init.zsh:6: is-at-least: function definition file not found
ERROR: Zim didn't start. You're using zsh version 5.4.1, and versions < 5.2 are not supported. Update your zsh.

```

- gist of .zimrc:
https://gist.github.com/9365c7fd0e077422823218aa2bdc8234
- gist of .zshrc:
https://gist.github.com/61fb47054d703434e6f663b02b324773
- gist of /etc/zsh/zshrc
https://gist.github.com/b348b7fb748932429082d52f62b92b96
- gist of init.zsh
https://gist.github.com/6b06766ac58cd0cebad6de69c38f6760

Thanks!

Eriner:
From `man zshbuiltins`:

with -R, an error message is printed and command processing aborted
immediately the search fails, i.e. at the autoload command rather
than at function execution..

Fixes #249

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -2,7 +2,7 @@
 # Zim initializition
 #
 
-autoload -Uz is-at-least
+autoload -UzR is-at-least
 if ! is-at-least 5.2; then
   print "ERROR: Zim didn't start. You're using zsh version ${ZSH_VERSION}, and versions < 5.2 are not supported. Update your zsh." >&2
   return 1


AtomicCoding:
@Eriner Why not just fall back to `echo "$ZSH_VERSION" | sed 's/\.//g'` if we can't use `is-at-least`?

Eriner:
@AtomicCoding Because if we're missing a `zshbuiltin`, the environment's fucked anyway. This is, in fact, an extremely convenient test for that, because if it fails, no zim code is executed at all and we don't leave the shell in a half-broken state.

If we can't rely on `is-at-least` to be present, then God only knows what else is broken and missing. The way this broke is interesting too due to the negative conditional, but that's a different matter.

ericbn:
Hey @Eriner 👋 

On `zsh 5.3 (x86_64-apple-darwin17.0)` I get

```
autoload: bad option: -R
```

And I don't see any documentation for that switch on `man zshbuiltins` either.

Eriner:
Arrrg. This wasn't added until 5.4.2: https://github.com/zsh-users/zsh/commit/f26d1ba6b01a358c83f28219c7a01e546e84d2ee

AtomicCoding:
[This](https://unix.stackexchange.com/questions/332005/test-for-functions-existence-that-can-work-on-both-bash-and-zsh) stack overflow question seems to have an answer.
Ripped from the answer:
```zsh
if typeset -f foo > /dev/null; then
  echo there is a foo function
fi
```

Eriner:
Yes, something like that will be necessary. It's unfortunate, but this is another case of nice features introduced in Zsh over a year ago that we can't use becasue we have to support old versions (due to the maintainers of those OS' and distros not updating packages on reasonable timelines -- par for the course.

Issue #138, opened 2017-02-12
Port "Sorin" theme


beloso:
I don't knot much about theming ZSH.

I would like to know if it's possible to port sorin's theme to zim?

testark:
@beloso não é para usar... :p

Eriner:
I'm not sure that it's a matter of "support" as much as it is someone taking the time to do the work and make the PR :)

beloso:
I'll try my hand at doing it. But I've never dived into ZSH themeing. might turn out messy

AtomicCoding:
Here's a prototype, it's messy, bad and not fully functional, but it's a starting point:
```zsh
#
# A simple theme that displays relevant, contextual information.
#
# Authors:
#   Sorin Ionescu <sorin.ionescu@gmail.com>
#
# Screenshots:
#   http://i.imgur.com/nrGV6pg.png
#

#
# 16 Terminal Colors
# -- ---------------
#  0 black
#  1 red
#  2 green
#  3 yellow
#  4 blue
#  5 magenta
#  6 cyan
#  7 white
#  8 bright black
#  9 bright red
# 10 bright green
# 11 bright yellow
# 12 bright blue
# 13 bright magenta
# 14 bright cyan
# 15 bright white
#

function prompt_sorin_async_callback {
  true
  # case $1 in
  #   prompt_sorin_async_git)
  #     # We can safely split on ':' because it isn't allowed in ref names.
  #     IFS=':' read _git_target _git_post_target <<<"$3"

  #     # The target actually contains 3 space separated possibilities, so we need to
  #     # make sure we grab the first one.
  #     _git_target=$(coalesce ${(@)${(z)_git_target}})

  #     if [[ -z "$_git_target" ]]; then
  #       # No git target detected, flush the git fragment and redisplay the prompt.
  #       if [[ -n "$_prompt_sorin_git" ]]; then
  #         _prompt_sorin_git=''
  #         zle && zle reset-prompt
  #       fi
  #     else
  #       # Git target detected, update the git fragment and redisplay the prompt.
  #       _prompt_sorin_git="${_git_target}${_git_post_target}"
  #       zle && zle reset-prompt
  #     fi
  #     ;;
  # esac
}

function prompt_sorin_async_git {
  cd -q "$1"
  if (( $+functions[git-info] )); then
    git-info
    print ${git_info[status]}
  fi
}

function prompt_sorin_precmd {
  setopt LOCAL_OPTIONS
[...]

PR #253, opened 2018-03-30
[Experimental] Sorin prompt

AtomicCoding:
Added Sorin Prompt from Prezto. The prompt depends on a LOT of Prezto functions so not everything works yet. Feel free to make changes!


AtomicCoding:
@Eriner @ericbn Can you review this?

ericbn:
@AtomicCoding, I have second thoughts about adding more prompt themes to zim. And it's [not even the first time](https://github.com/zimfw/zimfw/pull/233#issuecomment-355859662).  :- )

As [I've stated before](https://github.com/zimfw/zimfw/pull/192#issuecomment-331748148):

> I still think that the themes that come with zim, alongside our git-info module, provide a good basis for users to customize their own themes, and that we should not include variations of the already existing themes to our repo.

You can keep this as a fork under your GitHub user, if you intend to keep maintaining the code. Plus, the original sorin theme display many fine-grained pieces of information about git that our `git-info` module does not even support (because we want to keep things simple and fast).

AtomicCoding:
@ericbn Sorry, I made this in response to issue #138.

AtomicCoding:
@ericbn Should I close this?

PR #255, opened 2018-04-03
Remove old link to meta module in README

aoc:
Meta module was removed some time ago but the README wasn't updated to reflect that


ericbn:
Good catch. Thank you, @aoc!

PR #256, opened 2018-04-04
[input] Use custom expand-or-complete in zsh < 5.3

### LINKED ISSUE #134: Tabbing removes lines upwards
_First and foremost - thanks for developing a speedy zsh framework!_

I'm using the Arch AUR build **zsh-zim-git**, currently versioned `r269.61ff601-1`. I'm also currently running **zsh 5.3.1**.

*zimrc:* https://github.com/sQVe/susuwatari-dotfiles/blob/master/zimrc
*zshrc:* https://github.com/sQVe/susuwatari-dotfiles/blob/master/zshrc

Description
-----------
I've had this annoying problem for quite some time that I've been struggling with. When pressing <Tab> when trying to complete something that has **no matches** it will erase and move up one line in my terminal each time I press <Tab>.

Steps to reproduce
------------------

Type `idonotexist` and then <Tab> to try to complete it by pressing <Tab>.

Images or other information
---------------------------
screenfetch-2017-01-23_22-26-32
screenfetch-2017-01-23_22-26-48

ericbn:
as `expand-or-complete-with-redisplay` was added to fix an issue reported with zsh 5.2 when completing with <Tab><Tab><Ctrl-C>. See #85.

Fixes #134

### DIFF
--- a/modules/input/init.zsh
+++ b/modules/input/init.zsh
@@ -102,14 +102,16 @@ if [[ -n "${key_info[BackTab]}" ]]; then
   bindkey "${key_info[BackTab]}" reverse-menu-complete
 fi
 
-# Redisplay after completing, and avoid blank prompt after <Tab><Tab><Ctrl-C>
-expand-or-complete-with-redisplay() {
-  print -n '...'
-  zle expand-or-complete
-  zle redisplay
-}
-zle -N expand-or-complete-with-redisplay
-bindkey "${key_info[Control]}I" expand-or-complete-with-redisplay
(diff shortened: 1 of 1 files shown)


ericbn:
@davidzhuo1, please check if the changes in this PR fix your issue as you reported in #134.

@Eriner, do the changes here still keep #85 as a solved issue? I tested locally with `zsh 5.3` and was not able to reproduce it anymore.

Eriner:
LGTM. Might also be worth making a follow-up PR to implement the `is-at-least` check in other places (like [here](https://github.com/zimfw/zimfw/blob/master/init.zsh#L6))

Issue #258, opened 2018-04-05
Fish-style history search doesn't work in vi mode


antonshiryaev:
If I have `set -o vi` in `.zshrc`, it doesn't work. If I comment it out, vi mode is disabled, obviously, but it starts to work again.

antonshiryaev:
[Same issue, apparently](https://github.com/robbyrussell/oh-my-zsh/issues/1479#issuecomment-42633554)

AtomicCoding:
@Eriner @ericbn 
We could fix this with something like:
```zsh
if set -o | grep "vi                    on"; then
bindkey "something" history-substring-search-up
bindkey "something" history-substring-search-down
else
bindkey '^[[A' history-substring-search-up
bindkey '^[[B' history-substring-search-down
fi
```

AtomicCoding:
@antonshiryaev 
Does it work if you paste,
```zsh
bindkey "^[" history-substring-search-up
bindkey "^]" history-substring-search-down
```
into zsh and then try using `Ctrl+[` instead of up and `Ctrl+]` instead of down?

antonshiryaev:
In `.zshrc`, you mean. Moving `set -o vi` to the top of `.zshrc` and `history-substring-search` to the end of zmodules in `.zimrc` fixed it.

antonshiryaev:
Adding things before code loading the framework is not something many people would normally do - perhaps it'd be a good idea to set the mode through .zimrc and say so in the readme?

AtomicCoding:
@antonshiryaev PR? You can make your own changes to the README and zimrc or zshrc.

Issue #114, opened 2016-12-19
Shift+Tab doesn't cycle navigation backwards


antonshiryaev:
Not sure if it's a bug or an oversight, but shouldn't be too hard to fix either way.

Eriner:
Shift+tab cycles backwards for me in the menu expansion.

antonshiryaev:
[zshrc.zip](https://github.com/Eriner/zim/files/671951/zshrc.zip)
Here are my .inputrc and .zshrc files. I don't think much else could be the cause. I'm on ZSH 5.3.1, macOS 10.12.2, iTerm 3.0.12.

Eriner:
Are you still experiencing this issue?

antonshiryaev:
Yes, in fact. In fact, it's reproduced on a different machine with a mostly clean set up, so I suspect it's due to `set editing-mode viI` in my .inputrc. I've worked around it by adding `bindkey '^[[Z' reverse-menu-complete` in my .zshrc.

antonshiryaev:
Figured something out: if `set -o vi` goes after the aforementioned bind, shift+tab doesn't work. Apparently it's overwritten.

PR #261, opened 2018-04-05
Fix for a small bug with vi mode and history search

ferrum-salvator:
As per [this bug](https://github.com/zimfw/zimfw/issues/258), and possibly [this](https://github.com/zimfw/zimfw/issues/114), setting vi mode after initializing overwrites up and down arrow keys, disabling history substring search. The workaround was to move `history-substring-search` to the end of `zmodules` and  set the mode before loading zim in `.zshrc`, which is counterintuitive and, IMO, a bad habit. Instead, I've added a setting to `.zimrc` read by `init.zsh` that fixes that. I think prezto did something similar. Also modified readme to point out the new way to set editing mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Installing Zim is easy. If you have a different shell framework installed (like 
   source ${ZDOTDIR:-${HOME}}/.zlogin
   ```
 
-5. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file. To set vi mode, set `zinput_mode` in `.zimrc`.
+5. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file. To set input mode, set `zinput_mode` in `.zimrc`.
 
 Updating
 --------
--- a/init.zsh
+++ b/init.zsh
@@ -14,8 +14,8 @@ fi
 # Source user configuration
 [[ -s ${ZDOTDIR:-${HOME}}/.zimrc ]] && source ${ZDOTDIR:-${HOME}}/.zimrc
 
-# If necessary, set vi mode before loading modules
-if ${zinput_mode}='vi'; then
+# If necessary, set vi mode before loading modules to ensure history search with arrow keys works in vi mode
+if [ ${zinput_mode}='vi' ]; then
   set -o vi
 else
   set -o emacs
--- a/init.zsh
+++ b/init.zsh
@@ -15,7 +15,7 @@ fi
 [[ -s ${ZDOTDIR:-${HOME}}/.zimrc ]] && source ${ZDOTDIR:-${HOME}}/.zimrc
 
 # If necessary, set vi mode before loading modules
-if zinput_mode='vi'; then
+if ${zinput_mode}='vi'; then
   set -o vi
 else
   set -o emacs
--- a/init.zsh
+++ b/init.zsh
@@ -14,6 +14,11 @@ fi
 # Source user configuration
 [[ -s ${ZDOTDIR:-${HOME}}/.zimrc ]] && source ${ZDOTDIR:-${HOME}}/.zimrc
 
+if zinput_mode='emacs'; then
+  set -o emacs
+else
+  set -o vi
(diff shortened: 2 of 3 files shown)


init.zsh
```diff
@@ -14,6 +14,13 @@ fi
 # Source user configuration
 [[ -s ${ZDOTDIR:-${HOME}}/.zimrc ]] && source ${ZDOTDIR:-${HOME}}/.zimrc
 
+# If necessary, set vi mode before loading modules to ensure history search with arrow keys works in vi mode
+if [ ${zinput_mode}='vi' ]; then
+  set -o vi
+else
+  set -o emacs
+fi
+
```
ericbn: I think this should go to the `input` module.

README.md
```diff
@@ -65,7 +65,7 @@ Installing Zim is easy. If you have a different shell framework installed (like
   source ${ZDOTDIR:-${HOME}}/.zlogin
   ```
 
-5. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file.
+5. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file. To set input mode, set `zinput_mode` in `.zimrc`.
```
ericbn: Not sure if this is so important to go to the main README.

templates/zimrc
```diff
@@ -57,6 +57,9 @@ ztermtitle='%n@%m:%~'
 # Input
 #
 
+# Set to vi or emacs
+zinput_mode='emacs'
+
```
ericbn: I know we've been using configuration variables like this so far, but I prefer to start using zstyle ones from now on (and migrate the existing ones to zstyle format when we release version 1.0).

So here it would be something like `zstyle ':zim:input' mode 'emacs'`

@Eriner, thoughts?

templates/zimrc
```diff
@@ -12,7 +12,7 @@
 # The second line of modules may depend on options set by modules in the first
 # line. These dependencies are noted on the respective module's README.md.
 zmodules=(directory environment git git-info history input utility custom \
-          syntax-highlighting history-substring-search prompt completion)
+          syntax-highlighting prompt completion history-substring-search)
```
ericbn: Why the order change? `history-substring-search` preferably should not be before `input`, and `completion` should be the last.

antonshiryaev: It was suggested as part of the fix. It seems to work fine on my machine with it in the usual place, though.

README.md
antonshiryaev: I do think it should be mentioned at all, since it's a deviation from pure zsh settings, but it's not important where.

init.zsh
antonshiryaev: The entire point of the fix is that it has to be set before loading modules. If you could discover why that is the case, we could choose a better place for it, but I don't know enough ZSH to do so at the moment.

templates/zimrc
Eriner: I have no strong feelings one way or another, provided they don't become overly-verbose abominations like I've seen elsewhere. The entire point to the simple variables is just that; simplicity. That being said, I think this helps to clean up some pollution and makes the varibles more structured and accessible.

templates/zimrc
Eriner: We need to figure out some way to enforce module loading order to some degree, or alternatively introduce some "bootstrap" and "post-init" conditionals prior to module loading to handle some of these cases (editor mode, completions, etc).

antonshiryaev: Sure, a generic solution would be much better, but this is a start.

init.zsh
ericbn: Agreed. And I think the only place where there should be a documentation about this is here (not in the main README file).

templates/zimrc
ericbn: Agreed. I think it's possible to make the modules independent of each other, so they can be loaded in any order. The changes proposed here adhere to this idea.

Side comment: Maybe we should move the `compinit` code to `init.zsh` also, to guarantee the same for the `completion` module.

PR #262, opened 2018-04-06
Add Ruby Module.

cowboyd:
Ruby is a popular language, and has a number of tools in its ecosystem which support it.

RVM
-----
The most popular tool is RVM which lets you manage a multiplicity of ruby interperters on your system in the event that you have a bunch of different projects each requiring their own interpreter. If you have RVM installed, this module sets it up for you.

Bundler
------
Lots of handly aliases for working with bundler which, is basically a requirement for doing any sort of Ruby development.

### DIFF
--- a/modules/ruby/init.zsh
+++ b/modules/ruby/init.zsh
@@ -1,0 +1,26 @@
+# Load RVM into the shell session.
+if [[ -s "$HOME/.rvm/scripts/rvm" ]]; then
+  # Unset AUTO_NAME_DIRS since auto adding variable-stored paths to ~ list
+  # conflicts with RVM.
+  unsetopt AUTO_NAME_DIRS
+
+  # Source RVM.
+  source "$HOME/.rvm/scripts/rvm"
+fi
+
+#
+# Aliases
+#
+
+# General
+alias rb='ruby'
+
+# Bundler
+alias rbb='bundle'
+alias rbbc='bundle clean'
+alias rbbe='bundle exec'
+alias rbbi='bundle install --path vendor/bundle'
+alias rbbl='bundle list'
+alias rbbo='bundle open'
+alias rbbp='bundle package'
+alias rbbu='bundle update'
--- a/modules/ruby/README.md
+++ b/modules/ruby/README.md
@@ -1,0 +1,29 @@
+Ruby
+===
+
+Provides RVM integration and nice aliases for Ruby
+
+RVM
+---
+
+If you have RVM installed, this will automatically set it up without
+any required configuration to your dot files. To install RVM, see
+http://rvm.io
+
+Aliases
+-------
+
+### Ruby
+
+  - `rb` is short for `ruby`.
+
+### Bundler
+
+  - `rbb` bundle install
+  - `rbbc` bundle clean
+  - `rbbe` bundle exec
+  - `rbbi` bundle install --path vendor/bundle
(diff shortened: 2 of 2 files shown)


modules/ruby/init.zsh
```diff
@@ -0,0 +1,26 @@
+# Load RVM into the shell session.
+if [[ -s "$HOME/.rvm/scripts/rvm" ]]; then
```
AtomicCoding: Please use `${ZDOTDIR:-${HOME}}` instead of $HOME.

modules/ruby/init.zsh
```diff
@@ -0,0 +1,26 @@
...
+  # Unset AUTO_NAME_DIRS since auto adding variable-stored paths to ~ list
+  # conflicts with RVM.
+  unsetopt AUTO_NAME_DIRS
+
+  # Source RVM.
+  source "$HOME/.rvm/scripts/rvm"
```
AtomicCoding: Please use `${ZDOTDIR:-${HOME}}` instead of $HOME. (Again)

AtomicCoding:
@ericbn @Eriner This seems ready for merge. Thoughts?

Eriner:
I'm not opposed to the content, but we do need to work out the loading mechanism first, and I don't know if this belongs in the main repo. Most of the modules are utilities and convenience functions that impact almost all users (prompt, archive functions, git). Some of the modules in the repo (pacman, fasd, etc) I would ultimately like to remove from the core repo and move them into their own repos, providing we have some way of loading external modules. The PR looks good, but we need some more discussion and implementation around the ideas mentioned above before we decide how to handle this.

singularitti:
@Eriner Does that mean we need to start repos like `homebrew` on our own and request the organization `zimfw` to import it?

ericbn:
@cowboyd, we're moving to a [new plugin mechanism](https://github.com/zimfw/zimfw/pull/323) (currently on the [develop branch](https://github.com/zimfw/zimfw/tree/develop)) where modules are separate repos. Are you willing to maintain a ruby module in the zimfw organization (in a GitHub zimfw/ruby repository)?

bgrrtt:
I created https://github.com/bgrrtt/zimfw-rbenv, over the weekend. I don't use Ruby regularly, but needed to get it configured for a current project. What timing!? The module is rough, but it works for me on macOS. There are some strong arguments for rbenv over rvm. It seems like rvm, often, just works, but is heavy on the system, where rbenv is by a [guy](https://github.com/sstephenson) at the same company as the [guy](https://github.com/dhh) that wrote Ruby on Rails, https://basecamp.com/about.

ericbn:
Great! Let's move that to a zimfw repo and I'll make you maintainer of that repo. Should it be `zimfw/ruby` (since I see it's not rbenv specific, as is also creates aliases for bundler)?

bgrrtt:
I think one zimfw/ruby is the way to go!

Then something like `zstyle ':zim:ruby' version-management-tool {rbenv|rvm|homebrew|system}` to define which ruby version management tool they intend to use. 

Sound good! Sending your way!

ericbn:
@bgrrtt, I've sent you an invite to the organization. You should be able to transfer your repo after you accept the invitation.

ericbn:
Closing this in favor of https://github.com/zimfw/ruby

PR #263, opened 2018-04-07
[Completion] Add custom SPROMPT

AtomicCoding:
Sets colored SPROMPT by default and allows for user override.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -26,7 +26,7 @@ if [[ ! ${TERM} = 'dumb' ]]; then
 	if (( ${+zsprompt} )); then
 		SPROMPT=$zsprompt
   else
-  	SPROMPT='zsh: correct %F{1}%R%f to %F{2}%r%f [nyae]? '
+  	SPROMPT='zsh: correct %F{red}%R%f to %F{green}%r%f [nyae]? '
 	fi
 fi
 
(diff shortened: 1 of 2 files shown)


AtomicCoding:
@ericbn @Eriner Thoughts?

ericbn:
Thanks for your suggestion, @AtomicCoding. I prefer just going with

    SPROMPT='zsh: correct %F{red}%R%f to %F{green}%r%f [nyae]? '

and having no extra configuration. (And using color names instead of numbers)

AtomicCoding:
@ericbn New commit. Changed to an override like method so the user doesn't have to deal with manually overriding SPROMPT after completion is loaded.

ericbn:
Not sure I like this approach of having an extra config var.

Unfortunately checking with `(( ! ${+SPROMPT} )) && SPROMPT='... '` does not work as the var is already pre-set by zsh.

AtomicCoding:
@Eriner Thoughts?

AtomicCoding:
@ericbn Changed to `if [ ${SPROMPT} = '(Default Config)' ]; then SPROMPT='(Color Config)'; fi`. Thoughts?

ericbn:
My 2 cents: Simpler is better I think. These 3 lines of code could be just 1 line in the `.zimrc` file. And the user would get an "unexpected" behavior if he tries to set the default before loading zim.

AtomicCoding:
@ericbn 
Got it! Moved SPROMPT setting to zimrc.
Looks like this:
```zsh
# Set custom completion prompt
SPROMPT="zsh: correct %F{red}%R%f to %F{green}%r%f [nyae]? "
```

ericbn:
Merged with changes (this is a spelling correction prompt, not a "completion prompt").

PR #264, opened 2018-04-11
[Completion] Setopt GLOB_COMPLETE

### LINKED ISSUE #212: Tab completion incorrectly completes the wrong file when starting with the end of the filename

In a terminal, run `zmanage info` and paste the output below:

```
$ zmanage info
zsh: command not found: zmanage
```
But I'm running the zim master via Arch Linux AUR's zsh-zim-git: 9eaba9a

- gist of .zimrc: `zmodules=(completion)`
- gist of .zshrc: `source /usr/lib/zim/init.zsh`

Description
-----------
Tab completion incorrectly fully completes the wrong file when starting with the end of the filename.

Steps to reproduce
------------------
Tab completion works fine if starting from the start of the filename.
```
$ touch foo-bar foo-baz-bar
$ touch f<tab>
$ touch foo-ba<tab>
 -- file --
foo-bar      foo-baz-bar
```

However, if I start typing from the back (or middle) of the filename, it fails.
```
$ touch bar<tab>
$ touch foo-bar<tab>
           ^ cursor here
$ touch foo-bar
                ^ cursor here
```
I would expect the second tab to bring up the completion-selection menu. However, it immediately completes to the shorter version instead.

Another slightly different case is if I type the *entirety* of the shared string.
```
$ touch -bar<tab>
$ touch foo-bar
                ^ cursor here
```
In this case, I'm not even given the cursor in the middle, it jumps straight to the end.

AtomicCoding:
[asciicast](https://asciinema.org/a/O6q9zYTMtPwH8vfsfBt19GFje)
Fixes #212.
```
GLOB_COMPLETE
When the current word has a glob pattern, do not insert all the words resulting from the expansion but generate matches as for completion and cycle through them like MENU_COMPLETE.
The matches are generated as if a '*' was added to the end of the word, or inserted at the cursor when COMPLETE_IN_WORD is set. 
This actually uses pattern matching, not globbing, so it works not only for files but for any completion, such as options, user names, etc.
Note that when the pattern matcher is used, matching control (for example, case-insensitive or anchored matching) cannot be used. 
This limitation only applies when the current word contains a pattern; simply turning on the GLOB_COMPLETE option does not have this effect.
```

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -30,6 +30,9 @@ source ${0:h}/compdefs.zsh
 # the cursor is moved to the end of the word
 setopt ALWAYS_TO_END
 
+# Glob completions
+setopt GLOB_COMPLETE
+
 # Perform a path search even on command names with slashes in them.
 setopt PATH_DIRS
 


Eriner:
LGTM :+1:

Eriner:
Note that this does _not_ break case insensitivity on case-sensitive filesystems (linux, ext4), this just prevents us from making completion changes that _would_ make it case case-sensitive, which is dumb and we would never do anyway.

AtomicCoding:
@Eriner Oh, ok, the documentation wasn't too clear on that.

ericbn:
I think I have a [better solution](https://github.com/zimfw/zimfw/issues/212#issuecomment-380489820) for the #212 issue.

AtomicCoding:
@ericbn Yeah, I had already suspected this would be the case. Should I close the PR?

AtomicCoding:
Closed due to this [better solution](https://github.com/zimfw/zimfw/issues/212#issuecomment-380489820) that does not break completion globs.

PR #268, opened 2018-04-25
Unify the `for` syntax

ericbn:
For short single commands, prefer a one-liner `for` with the zsh syntax:
```
for x (foo bar) print ${x}
```
Otherwise just place `; do` on the same line as the POSIX `for ... in`:
```
for x in foo bar; do
  print ${x}
done
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Installing Zim is easy. If you have a different shell framework installed (like 
 2. Paste this into your terminal to prepend the initialization templates to your configs:
   ```
   setopt EXTENDED_GLOB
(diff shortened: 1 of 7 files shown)


ericbn:
Also, I've already [updated the Wiki](https://github.com/zimfw/zimfw/wiki/Code-Style-Guide/_compare/392533d0ca0f6e6eff92ff465b3b2496c1017963...b724a02cb51b46d5243ac59fb26cd5d2d286f5aa) accordingly.

@Eriner, do you agree with this?

Issue #232, opened 2017-12-01
New zsh session throws "fasd: function definition file not found"


yeongsheng-tan:
Please check the existing issues to make sure you're not duplicating a report.

For bug reports, please provide the following information:

In a terminal, run `zmanage info` and paste the output below:

```
Zim commit ref:   abbeeb9
Zsh version:      zsh 5.4.2 (x86_64-apple-darwin16.7.0)
System info:      Darwin YS-MacBook-Pro.local 17.2.0 Darwin Kernel Version 17.2.0: Fri Sep 29 18:27:05 PDT 2017; root:xnu-4570.20.62~3/RELEASE_X86_64 x86_64
```

- gist of .zimrc:
```
#################
# CORE SETTINGS #
#################
#
# Zim settings
#
# Select what modules you would like enabled.
# The second line of modules may depend on options set by modules in the first line.
# These dependencies are noted on the respective module's README.md.
zmodules=(directory environment git git-info archive debug ssh history input utility meta custom \
          syntax-highlighting history-substring-search prompt spectrum completion fasd)

###################
# MODULE SETTINGS #
###################

#
# Prompt
#

# Set your desired prompt here
zprompt_theme='gitster'

#
# Completion
#

# set an optional host-specific filename for the completion cache file
# if none is provided, the default '.zcompdump' is used.
zcompdump_file=".zcompdump-${HOST}-${ZSH_VERSION}"

#
# Utility
#

# Uncomment to enable command correction prompts
# See: http://zsh.sourceforge.net/Doc/Release/Options.html#Input_002fOutput
setopt CORRECT

#
# Input
#

# Uncomment to enable double-dot expansion.
# This appends '../' to your input for each '.' you type after an initial '..'
zdouble_dot_expand='true'

#
# Syntax-Highlighting
#

# This determines what highlighters will be used with the syntax-highlighting module.
# Documentation of the highlighters can be found here:
#   https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md
# For (u)rxvt, termite and gnome-terminal users,
# removing the 'cursor' highlighter will fix the disappearing cursor problem
zhighlighters=(main brackets cursor)


#
# SSH
#

# Load these ssh identities with the ssh module
zssh_ids=(id_rsa)

#
# Pacman
#

# Set (optional) pacman front-end.
#zpacman_frontend='powerpill'

# Load any helper scripts as defined here
#zpacman_helper=(aur)

# Customize to your needs...
[...]

Issue #186, opened 2017-05-26
Regression: Fasd broken on zsh 5.0.8 OSX 10.11.6 since recent commit


kiryph:
Reverting commit https://github.com/Eriner/zim/commit/c58f9015ec4f469fdf269f562d4098e67d7dccf9
resolves my issue:
```
❯ fasd_cd -d
fasd:383: no such file or directory:
```
This commits sounds completely unnecessary anyway.

ericbn:
The changes in that commit could have undesirable effects, as I [mentioned before](https://github.com/Eriner/zim/pull/180#issuecomment-301131889).

@kiryph, thanks for reporting this. Can you please verify if any of those environment variables that appear on that commit are set as empty in your environment?

kiryph:
No, none of them are defined: the command `env` returns:
```
TERM_SESSION_ID=w0t1p0:A4480DE5-2547-4DB3-9BB7-8B8C486BFC85
COLORFGBG=11;15
ITERM_PROFILE=Default
XPC_FLAGS=0x0
PWD=/Users/kiryph
SHELL=/bin/zsh
TERM_PROGRAM_VERSION=3.0.15
TERM_PROGRAM=iTerm.app
TERM=xterm-256color
HOME=/Users/kiryph
TMPDIR=/var/folders/d_/5kjxh9v92k752bvzf559fd3c0000gn/T/
USER=kiryph
XPC_SERVICE_NAME=0
LOGNAME=kiryph
__CF_USER_TEXT_ENCODING=0x1F5:0x0:0x0
ITERM_SESSION_ID=w0t1p0:A4480DE5-2547-4DB3-9BB7-8B8C486BFC85
SHLVL=1
OLDPWD=/Users/kiryph
PAGER=less
LSCOLORS=exfxcxdxbxGxDxabagacad
LS_COLORS=di=34:ln=35:so=32:pi=33:ex=31:bd=36;01:cd=33;01:su=31;40;07:sg=36;40;07:tw=32;40;07:ow=33;40;07:
GREP_COLOR=37;45
GREP_COLORS=mt=37;45
LESS_TERMCAP_mb=
LESS_TERMCAP_md=
LESS_TERMCAP_me=
LESS_TERMCAP_se=
LESS_TERMCAP_so=
LESS_TERMCAP_ue=
LESS_TERMCAP_us=
PROMPT_EOL_MARK=
VISUAL=vim
PYENV_SHELL=zsh
[...]

PR #270, opened 2018-05-07
[fasd] Move core code to external module

ericbn:
instead of we keeping and maintaining it, like we did in commits 126e950, 8f395be, c58f901.

On the other hand, we would lose custom changes we did, like merging clvv/fasd#75 and partially merging clvv/fasd#77. I think it's not our purpose to maintain our own fork of other projects.


ericbn:
@Eriner and fasd users out there, @deathbeam, @yeongsheng-tan, @kiryph, @justbur, @jalberto, what do you think?

kiryph:
>  I think it's not our purpose to maintain our own fork of other projects.

I would agree in general.

However, the original fasd on https://github.com/clvv/fasd is unmaintained. I still use fasd and would appreciate not to lose this functionality.

Googling for a maintained fork brought following result:

The project https://github.com/euank/pazi (a fasd alternative written in rust) suggests:

https://github.com/whjvenyl/fasd

Whjenvyl has disabled the issue tracker. Only PR's are possible. I am not sure if I would call it better maintained since no discussion is possible.

If someone knows a maintained fork where you can ask the maintainer for some level of feedback, please add it to this discussion.

I see a practical reason to just keep it as it is: it works, no fixes necessary and code reorganization is also workload.

Looking at the past maintenance workload for fasd on zimfw, fasd does not look too bad.

There have been
- 5 commits with the search word 'fasd' (1 of 5 introduces functionality, 2 of 5 Use zsh style tests for empty vars +1 regression, and two fixes (2 of 5))
- 4 issues with fasd in the title (#232, #186, #111, and #24 (feature request for fasd) )
- 2 pull requests with fasd in the title (#133 and #132 addressing the same issue)

The feature was introduced in March 2016 and there have been only two fixes necessary.
No outstanding issue and PRs.

For me everything works right now. The suggested change which might introduce regression and more work than just keeping it as it is. If there would be outstanding bugs which require more maintenance in zimfw, I would understand this. But right now this looks like premature cleanup just for the sake of good coding practice.

jalberto:
As a user, current implementation works as expected and caused zero problems for me since implemented.
As a dev, I understand the constant seeking for perfection and proper code structure :)

An alternative would be to create a wrapper for fzf (https://github.com/junegunn/fzf) to behave in similar fashion to fasd

Eriner:
FWIW I've stopped using fasd and am now using fzf.

PR #272, opened 2018-05-09
Fix >| symbol

mul14:
The `>|` symbol not renders correctly because it's inside of a table. 
> Notice table in markdown also use `|` symbol.


ericbn:
A similar escaping was removed by @Eriner at 4cee50c2d65fd8eb43bc42fd5db401cf7183676b. This escaping was removed by a contributor at 9529245cbb78c16d4a8bfc946b788c5e9101c53c.

I think a better solution is to avoid having pipes inside tables. In this case maybe we can just have the first sentence:

    Disables overwrite existing files with `>`

In the utility/README.md file, maybe something like:

    `l` with `${PAGER}`

mul14:
I think use bulleted list is a good idea. We can view the `>|` correctly with/without markdown renderer.

PR #275, opened 2018-05-22
[Completion] Allow completion to be loaded at an arbitrary location

AtomicCoding:
Currently, if you move the location of completion to before utility, Zim doesn't set `list-colors` correctly. However, sometimes if you move the completion module to the end of the `zmodules` array, you sometimes end up with errors with compdefs (it doens't happen with Zim's native modules, but I still think we should have something in place).
So, if you need to move completion, you end up with a prompt like:
Before
However, with the location changed, you then get:
After

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -59,6 +59,14 @@ zstyle ':completion:*' verbose yes
 zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' '+r:|?=**'
 
 # directories
+# If LS_COLORS is set, just use it
+if (( ${+LS_COLORS} )); then
+  # Set completion colors to LS_COLORS
+  zstyle ':completion:*' list-colors $LS_COLORS
+else
+  # Fallback to default LS_COLORS
+  zstyle ':completion:*' list-colors 'di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43'
+fi
 zstyle ':completion:*:*:cd:*' tag-order local-directories directory-stack path-directories
 zstyle ':completion:*:*:cd:*:directory-stack' menu yes select
 zstyle ':completion:*:-tilde-:*' group-order 'named-directories' 'path-directories' 'expand'
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -59,7 +59,6 @@ zstyle ':completion:*' verbose yes
 zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' '+r:|?=**'
 
 # directories
-zstyle ':completion:*:default' list-colors ${(s.:.)LS_COLORS}
 zstyle ':completion:*:*:cd:*' tag-order local-directories directory-stack path-directories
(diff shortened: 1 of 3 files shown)


modules/utility/init.zsh
```diff
@@ -12,6 +12,9 @@ if (( terminfo[colors] >= 8 )); then
   # GNU colours are used by completion module for all OSTYPEs
   (( ! ${+LS_COLORS} )) && export LS_COLORS='di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43'
 
+  # Set completion colors to LS_COLORS
+  zstyle ':completion:*' list-colors $LS_COLORS
```
ericbn: This is already in the [completion module](https://github.com/zimfw/zimfw/blob/56b61e211d2e6abed919d31845fcb89fc6ba15fe/modules/completion/init.zsh#L62).

AtomicCoding: The reason for the move is to allow the `completion` module to be loaded at an arbitrary location and stop `compdef` errors.

ericbn: I like the idea of making modules independent from each other.

Which `compdef` errors are you referring to?

AtomicCoding: For example, the `osx` module that I have installed from `Oh My Zsh`, runs `compdef _man man-preview`.

ericbn: That should be fixed by sourcing the `osx` module **after** the `completion` one. As `completion` runs compinit, which defines the compdef function.

AtomicCoding: @ericbn I know. But I still think we should make the completion module independent.

ericbn:
@AtomicCoding, your changes would make it only work if the utility module is sourced, and on the other hand I think we would expect to see this if the completion module is sourced.

Why don't we add the following to completion instead?
```
if (( ${+LS_COLORS} )); then
  zstyle ':completion:*:default' list-colors ${(s.:.)LS_COLORS}
else
  zstyle ':completion:*:default' list-colors ${(s.:.)di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43}
fi
```

We would make completion not depend on the LS_COLORS that are set by utility this way, thus making the modules independent.

modules/completion/init.zsh
```diff
@@ -59,7 +59,14 @@ zstyle ':completion:*' verbose yes
...
 
 # directories
-zstyle ':completion:*:default' list-colors ${(s.:.)LS_COLORS}
+# If LS_COLORS is set, just use it
+if (( ${+LS_COLORS} )); then
+  # Set completion colors to LS_COLORS
+  zstyle ':completion:*' list-colors ${(s.:.)LS_COLORS}
+else
+  # Fallback to default LS_COLORS
+  zstyle ':completion:*' list-colors ${(s.:.)di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43}
```
ericbn: We should use `':completion:*:default'` instead of `':completion:*'`, [according to the manual](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#index-list_002dcolors_002c-completion-style). Also I think all the new comments are redundant.

PR #277, opened 2018-06-12
support for external modules

jguddas:
I'm not much of an expert but i gave moving the modules out of the main repo a try and it seems to be working :smile:.


jguddas:
@ericbn what do you think about this?

braineo:
I think this feature is cool, managing submodule is not an advantage of git.

tools/zim_update
```diff
@@ -2,10 +2,35 @@
...
 #
 
+echo "\nUpdating Zim"
 cd ${ZIM_HOME}
 
 # this is the cleanest way I know how to update a repository
 git remote update -p
 git merge --ff-only @\{u\}
 # and update the submodules
 git submodule update --init --recursive
```
kaznovac: this is unnecessary as there are no submodules left :)

jguddas: this is so people can still manage their modules using submodules if they so chose to.

ericbn:
@jguddas, I've implemented this in the new develop branch of Zim. See #323 and see https://github.com/zimfw/zimfw/tree/develop

ericbn:
This is done with the release of Zim 1.0.0.

PR #279, opened 2018-06-21
pacman: add yay to SUID handling AUR helpers

sebastianst:
Yay handles sudo calls themself, so add it to the list. There are probably other suid aware helpers, see the [Arch wiki page on aur helpers](https://wiki.archlinux.org/index.php/AUR_helpers).


AtomicCoding:
@ericbn @Eriner This looks ready to merge.

PR #287, opened 2018-07-07
Make installation step easier to follow and show mechanism..

Piping:
Currently the installation steps is a bit confusing and requires some digging of the code..


Plloi:
I think you've got some commit bleed here.

Piping:
Should I merge these commits into one?

Plloi:
@Piping  No, but it you looks like you added more to the pull than intended. (the additional sub modules and the changes to init.zsh and login_init.zsh)

Piping:
@Plloi  I see, let me make a new branch and clean it a bit!

PR #288, opened 2018-07-10
Make installation step easier to follow and show mechanism...

Piping:
Currently the installation steps is a bit confusing and requires some digging of the code.
In the update, it points out that where is custom configuration for modules, theme, etc..

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,34 +38,27 @@ Zim has many modules! Enable as many or as few as you'd like.
 Installation
 ------------
 Installing Zim is easy. If you have a different shell framework installed (like oh-my-zsh or prezto),
-*uninstall those first to prevent conflicts*. It can be installed manually by following the instructions below:
+**uninstall those first to prevent conflicts**. Then put following instructions at the end of your `~/.zshrc`:
 
-1. In a Zsh shell, clone the repository:
-  ```
-  git clone --recursive https://github.com/zimfw/zimfw.git ${ZDOTDIR:-${HOME}}/.zim
-  ```
+  ```bash
+  # Zim initializition
+  export ZIM_HOME="$HOME/.zim"
 
-2. Paste this into your terminal to prepend the initialization templates to your configs:
-  ```
-  setopt EXTENDED_GLOB
-  for template_file in ${ZDOTDIR:-${HOME}}/.zim/templates/*; do
-    user_file="${ZDOTDIR:-${HOME}}/.${template_file:t}"
-    touch ${user_file}
-    ( print -rn "$(<${template_file})$(<${user_file})" >! ${user_file} ) 2>/dev/null
-  done
-  ```
+  if [ ! -f "$ZIM_HOME/init.zsh" ]; then
+    echo "Installing zim"
+    git clone --recursive https://github.com/zimfw/zimfw.git $ZIM_HOME
+    cat $ZIM_HOME/templates/zlogin >> $HOME/.zlogin
+    source $HOME/.zlogin
+  fi
 
-3. Set Zsh as the default shell:
-  ```
-  chsh -s =zsh
+  zmodules=(git git-info prompt completion syntax-highlighting autosuggestions)
+  zprompt_theme='steeef'
+  zhighlighters=(main brackets cursor)
+
+  source $ZIM_HOME/init.zsh #make sure init after zmodules lists etcs..
   ```
 
-4. Open a new terminal and finish optimization (this is only needed once, hereafter it will happen upon desktop/tty login):
-  ```
-  source ${ZDOTDIR:-${HOME}}/.zlogin
-  ```
-
-5. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file.
+You're done! Enjoy your Blazing Fast Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file.
 
 Updating
 --------


Piping:
@Plloi Just updated new pull request!

bgrrtt:
I've had a long running personal issue with the README changing the default shell AFTER pasting the template setup script into your terminal. The setup script does not paste correctly in bash, which is more often default.

If you change the shell to zsh FIRST, the setup script (`setopt`) works just fine. ;-)

- https://github.com/zimfw/zimfw/pull/288 removes these steps and makes installation automatic. Win.
- There is only one step...in the normal place to configure zsh. Win.

README.md
```diff
@@ -38,34 +38,27 @@ Zim has many modules! Enable as many or as few as you'd like.
...
+    source $HOME/.zlogin
+  fi
 
-3. Set Zsh as the default shell:
-  ```
-  chsh -s =zsh
-  ```
+  zmodules=(git git-info prompt completion syntax-highlighting autosuggestions)
+  zprompt_theme='steeef'
+  zhighlighters=(main brackets cursor)
```
Plloi: These 3 config lines are meant to be in the .zimrc file...

bgrrtt: What if there was no .zimrc? Settings could be added to .zshrc, or if not present, default functionality should execute anyway. Just thinking out load!

Plloi: I would prefer to keep zim settings separate, if for no other reason, so it's cleaner to switch to another plugin system.

README.md
```diff
@@ -38,34 +38,27 @@ Zim has many modules! Enable as many or as few as you'd like.
...
-  for template_file in ${ZDOTDIR:-${HOME}}/.zim/templates/*; do
-    user_file="${ZDOTDIR:-${HOME}}/.${template_file:t}"
-    touch ${user_file}
-    ( print -rn "$(<${template_file})$(<${user_file})" >! ${user_file} ) 2>/dev/null
-  done
-  ```
+  if [ ! -f "$ZIM_HOME/init.zsh" ]; then
+    echo "Installing zim"
+    git clone --recursive https://github.com/zimfw/zimfw.git $ZIM_HOME
+    cat $ZIM_HOME/templates/zlogin >> $HOME/.zlogin
```
Plloi: I'm asuming adding commands for installing zim to the zshrc is for users propagating settings to multiple machines. If that the case, why update .zlogin on every installation? The user may end up with multiple sourcings of login_init.zsh.

Piping: First, the zim initialization steps inside the zshrc could definitely help to propagating settings.
`if [ ! -f "$ZIM_HOME/init.zsh" ];` would guarantee that sourcing login_init.zsh is only happened once.

README.md
Piping: Modification on `.zshrc` is always need anyway. Deleting one line of code is no difference from deleting a block of continuous code. 

I actually prefer not to touch `.zlogin` at all and in my own setup I write my own 'zmanage reload' command to reload the `cache` when I needed it since tmux always open a login shell on split, and it slows down prompt about `50ms` on old server machines.

README.md
```diff
@@ -38,34 +38,27 @@ Zim has many modules! Enable as many or as few as you'd like.
...
-2. Paste this into your terminal to prepend the initialization templates to your configs:
-  ```
-  setopt EXTENDED_GLOB
-  for template_file in ${ZDOTDIR:-${HOME}}/.zim/templates/*; do
-    user_file="${ZDOTDIR:-${HOME}}/.${template_file:t}"
-    touch ${user_file}
-    ( print -rn "$(<${template_file})$(<${user_file})" >! ${user_file} ) 2>/dev/null
-  done
-  ```
+  if [ ! -f "$ZIM_HOME/init.zsh" ]; then
```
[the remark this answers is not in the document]
Piping: What do you suggest then? I think the effect here after compiling the zshrc will be insignificant, probably under 1ms.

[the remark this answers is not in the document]
kaznovac: create an installation script (./tools/zim_install) from this code and suggest users in readme to bootstrap zim installation from it e.g. `sh -c "$(curl -fsSL https://raw.github.com/zimfw/zimfw/master/tools/zim_install)"`

ericbn:
I understand this PR is trying to make two points:
1. Installation should be easier.
2. We should not have a `.zimrc` file, and put all the configuration into `.zshrc` instead.

We're creating a new version of Zim with a "plugin mechanism" (see #323), already available to try in the `develop` branch, and I'd to address these two points in this new version.

For the easier installation, I created an installer. See https://github.com/zimfw/zimfw/tree/develop#installation. It can't get easier than that, can it?  :- )

About not having a `.zimrc` file, and having everything in the `.zshrc`, I think there's a good point to this idea, since it would be clearer where to do the plugins/themes configurations in terms of **after** or **before** they're sourced, as there would be only one place to do it -- the `.zshrc` file -- and then in relation to the line where Zim is sourced. See https://github.com/zimfw/zimfw/issues/324#issuecomment-453651818

On the other hand, the new Zim implementation of the "plugin mechanism" sources the `.zimrc` before `zimfw install` or `zimfw update`, so you can enable/disable modules and then install/update them in the same shell session. I'm thinking a good balance here could be having only Zim's "plugin mechanism" configuration in the `.zimrc` and all the rest in the `.zshrc` file.

Thoughts? @Eriner, what do you think?

ericbn:
I'm almost done with the develop branch (see #323), and the last step I still want to have discussed is related to this PR. As I commented above, I plan to move the modules-specific configuration into `.zshrc`, and  leave `.zimrc` with just the configuration if which modules are enabled and custom url/branches/etc. of each one (the `':zim' modules` and the `':zim:module' <module>` styles, I'll call it **"meta configuration"** in the rest of this comment). 

I think this brings two advantages:
* As all module-specific configuration is in `.zshrc`, we will have these steps:
   * set "pre-source" configurations, which is all we have in our current `.zimrc`
   * source Zim, which on its turn sources the modules
   * set "post-source" configurations, which is something we can't have in the `.zimrc` file, and that can cover the [zsh-history-substring-search](https://github.com/zsh-users/zsh-history-substring-search) module among others.
* We keep the "meta" configuration in the `.zimrc` file, which is something we want to re-source before installing of updating modules.

@Eriner, @Plloi, thoughts?

shizonic:
Just a side note: what about renaming .zimrc to .modulesrc or .zimmodulesrc if it just holds these configurations in future?

Plloi:
pre-source/post-source makes a really strong argument for moving those options into .zshrc.

While I would still prefer to keep it separate, I can appreciate the added flexibility. I'm ok with the change.

PR #292, opened 2018-09-10
[ssh] Update init.zsh

ericbn:
based on https://stackoverflow.com/a/48509425/2654518, which is based on http://rabexc.org/posts/pitfalls-of-ssh-agents

Current code has a few issues: depends on `SSH_AUTH_SOCK` and `SSH_AGENT_PID` env variables, which might not be available in every shell session; and tries to create a new socket for agent-forwarding by checking `SSH_AUTH_SOCKET` instead of `SSH_AUTH_SOCK`.

Also, it's safer to create the env file with 066 flags and in the user home directory.

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -7,40 +7,28 @@ if (( ! ${+commands[ssh-agent]} )); then
   return 1
 fi
 
-# use a sane temp dir; creating 1k ssh-* files in /tmp is crazy
-if [[ ${TMPDIR} ]]; then
-  local ssh_env=${TMPDIR}/ssh-agent.env
-  local ssh_sock=${TMPDIR}/ssh-agent.sock
-else
-  # create a sane tmp dir at /tmp/username
-  mkdir -p /tmp/${USER}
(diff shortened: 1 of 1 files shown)


ericbn:
@Eriner, what do you think about the changes I'm proposing here on the ssh module?

ericbn:
@Eriner, for `(( ))` you can omit the `$`. [Remember?](https://github.com/zimfw/zimfw/commit/abbeeb9d375560b0e10c782e64970f22ff0dce11#commitcomment-25941707)  :- )

It works even for `?` and `#`. I could not find much about it in the documentation, but looks like the `$` would first substitute the variable by it's value and then the arithmetic operation would follow, while without the `$` the arithmetic operation would be performed directly using the variable name. And `?` and `#` are both integers, which make using `(( ))` a better choice:
```
% print -rl -- ${(t)#}
integer-readonly-special
% print -rl -- ${(t)?}
integer-readonly-special
```

Eriner:
@ericbn I had completely forgotten about that. (Old?) variable reference habits die hard.

PR #294, opened 2018-09-17
Fix ssh completion for ~/.ssh/config

### LINKED ISSUE #46: ssh tab completion with host name
- Zim commit ref: 6cb9d54
- Zsh version: 5.2
- $(uname -mosr): Linux 4.5.1-1-ARCH x86_64 GNU/Linux

Arch Linux, zim installed via AUR PKGBUILD (zsh-zim-git)
## Description

prezto offers tab completion of ssh with host names. e.g. if there are hosts defined in `~/.ssh/config`, you can tab complete with these. This seems to be missing in zim.
## Steps to reproduce

Type in `$ ssh f<tab>`. Expect host `foo` to be offered as a completion.

Eriner:
Ref: https://www.zsh.org/mla/users/2015/msg00467.html

and `which _ssh_hosts`

Fixes #46

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -92,3 +92,7 @@ zstyle ':completion:*:history-words' menu yes
 # ignore multiple entries.
 zstyle ':completion:*:(rm|kill|diff):*' ignore-line other
 zstyle ':completion:*:rm:*' file-patterns '*:all-files'
+
+# fix ssh not completing with hosts from ~/.ssh/config
+zstyle -e ':completion:*:*:ssh:*:my-accounts' users-hosts \
+  '[[ -f ~/.ssh/config && $key = hosts ]] && key=my_hosts reply=()'


ericbn:
I see only results from `~/.ssh/config`, and not from `~/.ssh/known_hosts` or `/etc/hosts` anymore, with this change.

Eriner:
We have a choice here. We can either use this to fallthrough to the ~/.ssh/config entries exclusively, or we can use the default behavior of reading known_hosts and /etc/hosts.

Prezto uses a wild custom zstyle to read everything: https://github.com/sorin-ionescu/prezto/blob/fbcae356d30928802e8150f4beca7e43f39513e1/modules/completion/init.zsh#L109

I don't want to do that because it's ugly and unmanageable. When that shit breaks, no one will want to fix it. Instead, I think we should fall back to zsh's built-in parsing of the config. It isn't perfect, but it's built-in. While we have to do some funky things to force that code path to traverse, I think this is the easiest implementation of the original ask in #46.

In terms of it being either ( /etc/hosts + ~/.ssh/known_hosts ) or ~/.ssh/config if present, I opt for the latter. The number of times I've wanted to **use completion** ssh to a box that is in known_hosts but not the ~/.ssh/config is near zero. Even in use-cases where reading hosts and known_hosts could be useful, typically it's a (temporary) IP-not-DNS situation and I'll be using substring completion anyway.

I think this change results in the _expected_ behavior. Users can always override this by overriding the zstyle, as that is a benefit of using zstyles.

ericbn:
@Eriner, makes perfect sense. Thanks for the explanation!

Also, it's an elegant solution, making use what is already provided by `_ssh`.

Sorry for not merging this before, I'm not being an active user of ssh completion lately...

I've changed the code to match our styling, changed the comment so it explains what is in the original solution that you referenced (people can define a `_my_hosts` function to override the completion), and added your motivation for the change to the commit message.

PR #296, opened 2018-09-26
[pacman] Remove dependency from completion module

ericbn:
The completion module has specific code for the pacman module, and we want to make modules independent from each other.

Plus, as far as I checked, most of the pacman wrappers/frontends are already packaged with zsh completion out of the box. I'm not an archlinux user, but I checked the repos of the ones in https://wiki.archlinux.org/index.php/AUR_helpers#Pacman_wrappers and these contain zsh completion scripts:

* aura: https://github.com/aurapm/aura/blob/master/aura/doc/completions/_aura
* pacaur: https://github.com/rmarquis/pacaur/blob/master/zsh.completion
* pakku: https://github.com/kitsunyan/pakku/blob/master/completion/zsh.patch
* pikaur: https://github.com/actionless/pikaur/blob/master/packaging/usr/share/zsh/site-functions/_pikaur
* trizen: https://github.com/trizen/trizen/blob/master/zsh.completion
* yay: https://github.com/Jguer/yay/blob/master/completions/zsh

I didn't find zsh completions for the ones listed below, so I added a compdef function for them inside the pacman module itself:

* aurman
* packer-aur
* wrapaur
* yaourt


ericbn:
@Eriner, @sebastianst, @Tmplt and other archlinux users, can you please check if this makes sense?

The main goal is here is to make modules independent, and we can set the completions from inside the pacman module by using a compdef function, as I'm doing here.

sebastianst:
My `pacman` autocompletion (e.g. `pacman -Rs spo<tab>` completes to deinstall spotify) still works but weirdly, `yay`s doesn't. The same command should complete but it doesn't. What are the other Archers' experiences?

Tmplt:
I no longer use Arch, so I'm not able to help.

Eriner:
I don't have time to test the functionality of this at the moment, but I like the direction. Assuming it works as expected, using the built-in compef structure is far superior to my lazy implementation. No real file access changes either in this case. Nice!

sebastianst:
Thanks @ericbn , I can confirm that `rm -f .zcompdump*` followed by `compinit` fixes my completion cache and `yay` completion now works as `pacman`'s does. So I give this PR a go.

PR #297, opened 2018-09-29
do not track new files in submodules from main repository

kaznovac:
prevent main repository `git status` reporting submodules having new files
new files are usualy cimpiled zsh files
there is no way to .gitingore from main repository specific files in submodules

Please follow Zim's [code style guidelines](https://github.com/zimfw/zimfw/wiki/Code-Style-Guide).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,21 +1,28 @@
 [submodule "modules/history-substring-search/external"]
 	path = modules/history-substring-search/external
 	url = https://github.com/zsh-users/zsh-history-substring-search.git
+	ignore = untracked
 [submodule "modules/completion/external"]
 	path = modules/completion/external
 	url = https://github.com/zsh-users/zsh-completions.git
+	ignore = untracked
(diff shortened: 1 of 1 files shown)


ericbn:
All submodules I checked have a `.gitignore` with `*.zwc`, so this solves compiled files being reported there.

I think not reporting any untracked file can cause more confusion than do any good. What if a new file was created on purpose?

kaznovac:
currently neither of extenal-themes have `.gitignore` file (arguably this is responsibility for their repos)

submodules are checked-out in detached-head state; creating files in order to push them upstream is highly unusual, and requires some git 'magic', but it's still posible.

this setting turns off main git (zimfw's) tracking of new files in submodule directory (`git status` from `zim` folder will not be reporting new files in sub-repos) but git repositories of sub-repos continue to behave as expected.

Eriner:
I do not have an issue with this because these are "magic" dependencies which zim users don't directly interact with and even manage them through zim's updating facilities (`zmanage update`).

PR #302, opened 2018-10-17
Add pikaur to list of AUR helpers that handle SUID

hyshka:
`pikaur` (https://github.com/actionless/pikaur) handles sudo authentication itself.


ericbn:
@hyshka, thanks for the contribution! 

@Eriner, @sebastianst, can you please review? I'm not an Arch user.

Issue #306, opened 2018-11-02
archive multiple files


Konfekt:
Regarding 

https://github.com/zimfw/zimfw/blob/60b0910f6def220f1b115e3934088a6a21b538da/modules/archive/functions/archive#L1

it could be useful to be able to compress files instead of a single directory, have a look at

https://github.com/robbyrussell/oh-my-zsh/pull/6846/files

for a possible approach.

ericbn:
Good idea, @Konfekt. I can work on it, or I'll be happy to look at a PR with this change if you're willing to do it.

PR #308, opened 2018-11-04
archive: add support for archiving multiple files

Konfekt:
See https://github.com/zimfw/zimfw/issues/306

It is POSIX shell syntax.

### DIFF
--- a/modules/archive/functions/archive
+++ b/modules/archive/functions/archive
@@ -2,40 +2,47 @@
 # Creates archive files
 #
 
-if (( ${#} != 2 )); then
-  print "usage: ${0} [archive_name.ext] [/path/to/include/in/archive]" >&2
-  return 1
-fi
+archive() { # compress a file or folder
+  local name="${1%.[^.]*}"
+  local type="${1##*.}"
+  local input="$2"
 
-# we are quitting (above) if there are not exactly 2 vars,
-# so we don't need any argc check here.
+  if [ "$#" -eq 1 ]; then
+    echo "archive(): archive a file or directory."
+    echo "Usage:   archive <archive> <files>"
+    echo "Error: Please give a valid file or directory name!"
+    return 1
+  fi
 
-# strip the path, just in case one is provided for some reason
-local archive_name="${1:t}"
-# use absolute paths, and follow symlinks
-local dir_to_archive="${2}"
+  if [ -f "${name}.${type}" ]; then
+    echo "archive(): file ${name}.${type} already exists."
+    name=$(mktemp --dry-run "./${name}_XXX.${type}")
+    echo "archive(): using ${name}.${type} instead."
+  fi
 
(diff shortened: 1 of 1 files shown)


Konfekt:
Be aware that for example `tar.xz` does currently not support archiving directories, but only files.

ericbn:
You should respect the project's [code style guide] (https://github.com/zimfw/zimfw/wiki/Code-Style-Guide). Projects usually have one, with most parts of it explicitly documented in a style guide, but some more subtle ones implicitly visible in the code itself. Respecting that is respecting the developers that came before you...

Plloi:
@Konfekt here's what is I see style wise:
* Variables not always braced
* Case format altered
* Comments removed, not updated (personal peeve)

Konfekt:
Thank you. I updated this pull request accordingly.

ericbn:
@Konfekt, if you're making a contribution to a repo (specially if it's the first one), you should push just the specific change you're proposing. You proposed supporting multiple files as arguments for `archive`, but you came up with lots of other changes:

* Code is inclosed in `archive() {...}`. This is an autoloaded function file, so no need for that.
* You use `[...]`. We prefer `((...))` for arithmetic conditions and `[[...]]` for other condition evaluations.
* You changed the error messages. I prefer we keep them as they are.
* You're creating a file with a different name if the provided name already exists. I prefer we keep it as we're doing.
* You changed the logic of most of the tools by changing the parameters (e.g. you use `tar j`, we use `tar --use-compress-program=bzip2` because we alias pbzip2 to bzip2 if the tool is available, and we want to take advantage of that), or by the way the tools are called. Each one of these changes deserve a separate discussion.
* You're forcing predefined extensions (e.g. you force `"${name}.tar.gz"` even if the provided extension was `.tgz`). We don't do that, and I prefer we keep it this way.
* You added support for `.{tZ,tar.Z,Z,compress,lzo}` files. That's great, but should be in a separate pull request with a separate discussion, as I think we should also update `unarchive` and `README.md` accordingly.

One more note: you're not treating the `files` array correctly, as it's not being expanded as different parameters when the tools are called with `${files}` at the end. Oh, and the extra `shift`s you have inside `case...esac` are unnecessary, as you're already doing one in the beginning of the function.

Konfekt:
Thank you for your corrections. One final doubt:

> One more note: you're not treating the files array correctly, as it's not being expanded as different parameters when the tools are called with ${files}

Please help me understand why it is treated incorrectly. How would it be treated correctly?

Plloi:
> * You use `[...]`. We prefer `((...))` for arithmetic conditions and `[[...]]` for other condition evaluations.

I feel like this one should be added to the style guide

Konfekt:
Thank you!

ericbn:
@Plloi, good point. Just updated the [style guide](https://github.com/zimfw/zimfw/wiki/Code-Style-Guide).

@Konfekt, you're welcome!

PR #311, opened 2018-11-20
Use which to get zsh path in install instructions

ericbn:
Instead of plain `=zsh` (is the `=` a typo?)


ericbn:
@Eriner, I'd like your thoughts here too.

Eriner:
`=zsh` == `/usr/bin/zsh` == `$(which zsh)`. The only difference is the latter drops into a subshell and works in shells other than zsh.

ericbn:
Cool. Just learned one more! Closing this.

PR #312, opened 2018-11-22
[archive] Allow archives in other directories

ericbn:
Current code will create an archive in the current directory with:

    archive ../test.tar.gz test.*

or will complain that the following archive doesn't exist in the current directory (given it actually exists in the parent one):

    unarchive ../test.tar.gz

Other changes:

* Use `<required_param>` instead of `[required_param]` in the usage text
* Don't explictly check if archive exists in `unarchive`, but let the respective tool fail with its own message

### DIFF
--- a/modules/archive/functions/archive
+++ b/modules/archive/functions/archive
@@ -1,17 +1,17 @@
+# vim:et sts=2 sw=2 ft=zsh
 #
 # Creates archive files
 #
 
 if (( # < 2 )); then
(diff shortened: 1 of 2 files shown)


ericbn:
@Eriner, I'd like your thoughts on these changes for the archive module functions.

Eriner:
Yeah, this looks fine. Don't know why I enforced it to all happen in the current directory. If it turns out that there was some bug, I likely would have documented that in a code comment. LGTM

PR #313, opened 2018-12-08
[history] Remove EXTENDED_HISTORY and INC_APPEND_HISTORY

ericbn:
as they are not required (not even recommended) to be set along with `SHARE_HISTORY`. See zshoptions(1):

> This option ... also causes your typed commands to be appended to the history file (the latter is like specifying `INC_APPEND_HISTORY`, which should be turned off if this option is in effect). The history lines are also output with timestamps ala `EXTENDED_HISTORY` ...

Also update copy in comments and in the README.

Use `fc -l` instead of `history`, as `fc` is the actual Zsh builtin. Add the `-n` flag to suppress event numbers when listing.

### DIFF
--- a/modules/history/README.md
+++ b/modules/history/README.md
@@ -1,26 +1,20 @@
-History
+history
 =======
 
-Sets sane default history options.
+Sets sane history options.
 
-History file is set to save in `${ZDOTDIR:-${HOME}}/.zhistory`
+The history is set to be saved in the `${ZDOTDIR:-${HOME}}/.zhistory` file.
 
-(most likely ~/.zhistory)
-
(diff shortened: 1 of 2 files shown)


ericbn:
@Eriner, after experimenting with different styles for the README file, I want to stick with the one I'm using in this PR, that is:

* Header 1 with the module name is in `lowercase`. 
* Other headers are in `Sentence case`. Common header names that should be consistently used are `Aliases`, `Functions`, `Settings`, and `Zsh options`.
* The names `Zim` and `Zsh` always appear capitalized, even in the middle of sentences.

Let me know what do you think.

PR #281, opened 2018-06-26
Add `homebrew` module

singularitti:
Originally forked from project [Prezto](https://github.com/sorin-ionescu/prezto) `homebrew` [module](https://github.com/sorin-ionescu/prezto/tree/master/modules/homebrew). Thanks to [Sorin Ionescu](https://github.com/sorin-ionescu) for the very first contribution!

I am not an expert of zsh, but I tried to fix minor "un"zsh style of the code in the [original code](https://github.com/sorin-ionescu/prezto/blob/0d6e7de0312366322bf7e7e1dbb41697b5bcd3b1/modules/homebrew/init.zsh#L9).
This module is simple and I would like to maintain it since I am a macOS user.

Issue #88, opened 2016-09-28
Better manage Zim modules and prompts allowing the project to grow in an organized way


ericbn:
I believe Oh My Zsh and even Prezto failed to grow in an organized fashion mainly because they lack incorporating some kind of plugin system that would allow people to contribute with external modules and keep the core clean and simple.

I suggest we take this step with Zim (and make it awesome)!

My suggestion would go in the line of:
- Creating a `zim` GitHub organization, having the Zim core be `zim/zim`, and other Zim modules be like `zim/xyz`.
- Keeping into the Zim core only what we would agree are the basic modules.
- Allowing users to configure external prompts as easily as having this in `.zimrc`:

        zprompt_theme='subnixr/minimal'

- Allowing users to declare external modules also as easily as (two last ones):

        zmodules=(directory environment git history input utility meta custom \
                  syntax-highlighting history-substring-search prompt completion \
                  zim/pacman foo/bar)

Maybe we can incorporate an external existing plugin system into the Zim core to accomplish this, or we can implement our own system (which I think would be the more flexible way to go).

The plugin system would have at least functions like:
- `install` to install new declared modules
- `update` to update all existing modules
- `clean`/`cleanup` to remove unused modules that are still installed

In case we use an internal plugin system, that could be managed by `zmanage`, which we already have. (Side suggestion: wouldn't it be cooler to rename `zmanage` to `zim`?)

Maybe the `zim` organization could be the default for external modules, so if the user declares simply `pacman` inside `zmodules` and there's no such module, the plugin system would try to check for a `zim/pacman` repo to install from there.

I would not bother making the plugin system capable of importing Oh My Zsh or Prezto modules.

My suggestion of the basic modules for the Zim code are:
- All those importing the Zsh ones: `completion`, `history-substring-search`, `syntax-highlighting`
- All the ones configuring the sane Zsh defaults for all/most OSs: `directory`, `environment`, `history`, `input`, `spectrum`(?), `ssh` (only not useful for Mac OS X), `utility`
- The zsh core ones: `debug` and `meta`
- `git`, as git is needed to install and keep Zim and external modules updated
- `prompt` with a small selection of internal prompts (we have `eriner`, `gitster`, `magicmace`, `steeef`; there's already a newer fork of `minimal` as reported in https://github.com/Eriner/zim/issues/78 and we could check if that is the case with any of the others)

I'd suggest `archive`(?), `fasd` and `pacman` go as external modules inside the `zim` organization.

[...]

Issue #33, opened 2016-02-16
Add virtualenv module


alexschneider:
For working on python projects - it'd be nice to see the virtualenv currently running.

aplatest:
This would need some sort of a hook into the current prompt. The same hook could then be used for rbenv and maybe other frameworks.

Eriner:
Would say this could be closed by #250, but yes we should make our own to hook into `prompt`.

PR #323, opened 2019-01-07
[develop] Add a plugin mechanism

### LINKED ISSUE #299: [rfc] submodule update process can checkout latest commit
Description
-----------

some of the submodules are fixed on very old versions.

we can use `git submodule update --remote` when updating modules
it will checkout latest commit from specified branch in submodules config [master if none is specified]
more info can be found here: https://git-scm.com/book/en/v2/Git-Tools-Submodules#_pulling_in_upstream_changes

benefit is obvious, users will get latest available version of submodule

downside is that one can get into inconsistent state if there is some incompatibilities within modules or the selected branch is not stable.

i would suggest implementing two update functions [or a parameter to update command]
one that updates to zimfw repository stated submodule commits `zmanage update` [or `zmanage update --safe`]
and another to update to latest `zmanage update-latest` [`zmanage update --latest`]
and `zmanage info` should also list `git submodule status also`

p.s.
i am currently using all the latest versions without any noticeable problems - it solved me an error message from syntax-highlighter complaining about README.md not implementing some methods when using `sudo -s` on macos.

ericbn:
This is a major change, where Zsh modules/plugins are not git submodules in the Zim repo anymore, but customized and installed separately as individual repositories. The discussion about this started more than 2 years ago in #88. Closes #277, closes #299.

This will allow contributors' modules to live in their own repositories.  See #33, #138, #262, #281, #324.

Some discussion topics that I think are worth considering before merging this:
- [x] Reduce the Zim "core" to a single file
- [x] Simplify installation with an installation script (See #182)
- [x] Put the configuration into `.zshrc` instead of a separate `.zimrc` (See #288)
- [x] Rerun the [zsh-framework-benchmark](https://github.com/zimfw/zsh-framework-benchmark)
- [x] Do not support themes that require `promptinit` (See #325)
- [x] Generate a static script that autoloads the functions and sources the modules

I suggest we create individual GitHub issues/PRs to start the separate discussions.

The current code has what, up to this point, I considered to be the best balance between simplicity, execution speed and number of files.

One measured decision was to make the initialization of modules depend only on the `':zim' modules` style, keeping it as fast as possible.  The `':zim:module' module` style is used to install, update and clean the modules, all operations that happen after the user got his as-blazing-fast-possible shell prompt.

Even though I didn't care much about making install or update fast, `xargs` has a nice feature of allowing commands to be executed in parallel with `-P`. I took advantage of that.

I've also worked on making the `zimfw` utility give the user some nice (while still minimalistic) output. Also I'm suggesting this as the new name for the `zmanage` tool, since `zimfw` does not shadow the `zim` wiki tool.

I strongly recommend you install this from scratch in a separate directory, instead of checking out `develop` in your current Zim installation repo.


ericbn:
Changes about prompt themes I forgot to mention: 
  * I've made each one of them individual repos.
  * I've removed the external themes: lean, liquidprompt, and pure. Didn't even document how to install them back.
  * Renamed the `minimal` theme to [`s1ck94`](https://github.com/zimfw/s1ck94), in honor of it's original creator, and since it was deprecated in favor of another theme also called `minimal`.  @PatTheMav, you don't need to call your Zim version of `minimal` as `minimal2` anymore.  :- )

PatTheMav:
@ericbn Yay! 🥳 Will read up on the changes over the week to understand how the theme would need to be adapted. Thanks for the work!

ericbn:
Added an automatic installer. Now that we have one, we're even more free to discuss how Zim will be distributed ("Reduce the Zim "core" to a single file?"). The new installer closes #182.

Plloi:
@ericbn Nice, moved over to the develop branch as soon as i saw this, and it's much cleaner then what i hacked together for handling this.  I was also able to get @PatTheMav's minimal2 working as is once my zimrc was updated.

When i made the move, copied my existing zim install to a new folder checked out develop just to see what would happen. Had to blow out modules.  What's the upgrade plan once this makes it into master?

edwardsmit:
@ericbn Just wanted to let you know I'm using this and am very happy with it. Adding the powerlevel9k prompt was super-easy.

edwardsmit:
@ericbn Very minor point: I think `modules/` should be added to `.gitignore`

ericbn:
I've just updated the zsh-framework-benchmark code and ran it. I included the Zim develop branch for comparison. See https://github.com/Eriner/zsh-framework-benchmark/pull/6

This is the results I got:

Framework | Mean (s) | Min…Max (s)
--- | --- | ---
zplug | 0.7029 ± 0.00682569 | 0.69…0.74
oh-my-zsh | 0.3323 ± 0.00443959 | 0.33…0.35
prezto | 0.2702 ± 0.00282135 | 0.26…0.29
zim | 0.1 ± 6.71586e-09 | 0.10…0.10
zim-develop | 0.0972 ± 0.00470744 | 0.09…0.11

Zim is still the top performer, and the develop and master branch keep the same performance level.

ericbn:
I'm adding a 5th point to the list we have in the description: should we support themes that require `promptinit`? My idea is that we shouldn't, see #325.

PatTheMav:
@ericbn Tried out the develop branch and the install script. Works fine so far, but the install script overwrites my Dotfile symlinks for .zshrc et. al. with prepended copies of the symlinked files. Might be an edge case but just FYI.

kaznovac:
Great job @ericbn

just tested it, with install script installing is as easy as it gets.
minimal theme is great addition, love it.

some suggestions:
- _templates_ folder is more _install_ repo related than _zim-core_, although that move will require install script to download templates for the cause of _zim-core_ being clean (there is no reference to template files from core)
- not a fan of utf8 characters in scripts ('✓' and '✗'), non-ascii characters can render incorrectly on machines with incorrect locale settings
- drop the `promptinit` definitely
- _single file core_ can be a good choice now (i'm thinking self-updating script file, no need for git overhead for _core_; or just call _install_ script to update the _core_)

templates/zimrc
```diff
@@ -1,96 +1,107 @@
...
-#   http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Simple-Prompt-Escapes
-# The example below uses the following format: 'username@host:/current/directory'
-ztermtitle='%n@%m:%~'
+# Enable double-dot expansion.
+# This appends '../' to your input for each '.' you type after an initial '..'
+#zstyle ':zim:input' double-dot-expand yes
 
 #
-# Input
+# pacman
```
kaznovac: _pacman_ and _ssh_ sections are 'dangling' - these two modules are not in the default module selection

Eriner:
Haven't had time to review this, but @ericbn please ping me when it's ready for a final code review - great work!!

ericbn:
Just pushed breaking changes (that's why I force-pushed). **Now the Zim "core" is a single file!** Also updated the automated install script accordingly. Please reinstall and enjoy the newest improvements.

@edwardsmit, this solves the problem of having ZIM_HOME as a git repo, and needing to "gitignore" the modules inside it. Now the zimfw repo is not cloned anymore, but just the single-file `zimfw.zsh` script is downloaded using either wget or curl and it's all in there.

ericbn:
Last step to consider this done is to move the modules-specific configuration into `.zshrc`, and  leave `.zimrc` with just the configuration if which modules are enabled and custom url/branches/etc. of each one (the `':zim' modules` and the `':zim:module' <module>` styles). Want to discuss this in #288.

PatTheMav:
@ericbn I have a bit of a hard time getting `zimfw update` and the other management commands to work via `make` (I use a makefile in my dot files to update my environment every day), because of the way the `zimfw` command is structured right now (it looks like you're creating a `ztool` variable on the fly that contains the subroutine code as shell commands).

In essence, the makefile route needs a way to call `zimfw update` from a `sh` session. Before your last changes, I could just run `zsh zim-update.zsh` which would contain a copy of the old update code, but I'm not happy about the code duplication.

ericbn:
@PatTheMav, you can call `zimfw` using a Zsh interactive session with:

    zsh -ic 'zimfw update'

It's fair enough to assume you have `zsh` installed if you want to call `zimfw`.  :- )

PatTheMav:
@ericbn Yeah `zsh` is a prerequisite for all `zim` targets, so that will be ensured. The `ic` switches did the trick, thanks.

ericbn:
@PatTheMav, thanks to you for helping beta test this!  :- )

PatTheMav:
@ericbn Sure thing - been running this on all my systems since you created the PR. Was a bit rocky at times, but that was due to the switch between versions.

Whenever the develop branch becomes active, special care should be given to pre-existing installations, as `.zshrc`, `.zlogin` and `.zimrc` require changes and the automatic prepending of the install script might break things (as there can't be an update script without unreasonable support burden).

ericbn:
@PatTheMav, yeah I'm aware we'll need a plan before promoting this branch to make the transition as smooth as possible for existing users. We did something in these lines, but waaay much simpler when we deprecated support for Zsh versions < 5.2 back in #184.

ericbn:
Something else I think is worth considering is generating a static script that is sourced by `.zshrc`, instead of sourcing a dynamic script that iterates the modules and figures out what to do on the fly, as we do. There would be no much performance gain (I observed a 2.3% reduction of the time for the default installation of Zim on Zsh 5.7.1). But it opens the possibility of not having to solely rely on the modules zstyle and knowing only which modules are enabled (with other configurations currently only affecting how modules are installed, but not how they are initialized). And any changes in more complex directions would not affect the initialization performance.

PatTheMav:
Maybe run `m4` or something (no Ruby requirement) to compile a launch-script from a given configuration using templates and a given config. Then source the generated file if it exists. File is generated using the installation and update commands respectively. That way the plugin structure can stay, but you still "just" source a script.

It kind of reminds me of your efforts to combine all modules into a single `zcompile`d file though. Didn't you do benchmarks back then as well?

jkaan:
Hello there! This might not be the right place to ask but I was wondering. Is it possible for me already to update/transfer to this perhaps unstable version of Zim? 

Are there any caveats that are good to know?

ericbn:
@jkaan, I would say the current develop code works really well. No known issues. The only caveat is that most probably there will be breaking changes in the future. For example, I'm still considering:

* Removing the directory, environment and history modules, in favor or moving the respective configuration to the zshrc template (the "set the default pager" code goes to the utility module, and the termtitle code to a separate termtitle module). See https://github.com/zimfw/environment/pull/1.
* Having a build command that creates a static initialization script, instead of dynamically initializing modules every time we start a new shell.
* Because of the previous item, also improving the module configuration options, allowing to configure the fpath, autoloaded functions, and sourced scripts per module. This might call for redoing the zstyle configuration format too.

PatTheMav:
@jkaan I've been using the develop branch more or less since @ericbn made it public here and haven't had any issues except for a bit of config snafus during the transition. If if were up to me this could already be the mainline version, but the points @ericbn mentioned need to be decided/implemented before that makes sense.

But `zimfw` is in a decent-enough place right now that some early momentum has slowed done a little.. 😅

edwardsmit:
@ericbn Can we help you with anything to get this merged?

Eriner:
Sorry I haven't had any time to give this a thorough review - just noting that https://github.com/zimfw/zimfw/commit/954f53d01d05a9841e96ae0f044bbb3f92d696e5 will need to be cherry-picked for merge.

ericbn:
The current develop code introduces breaking changes in relation to the master branch code, which will mostly require Zim users to do a fresh installation (or we need to think of a smooth migration process before merging). 

I'm still working on [items I mentioned earlier](https://github.com/zimfw/zimfw/pull/323#issuecomment-517938454). These items still introduce further breaking changes to the actual code we have in the develop branch, and I don't want to disrupt Zim users twice...

Eriner:
> Having a build command that creates a static initialization script, instead of dynamically initializing modules every time we start a new shell.

@ericbn This is something I've given a lot of thought too. In the perfect world, we pseudo-compile the code into a single `.zshrc` file. Calling only a single fopen would be a relatively massive speed increase. Every time I think about this though, it makes me want to "abandon" this project and create a glorified templating engine in Go that gobbles everything up and spits out a single `.zshrc`.

PatTheMav:
Compiling everything into one file is probably the biggest speed boost you can get, but I think that the current state of the develop branch that allows full compatibility with other `zsh` plugins is a big enhancement. Also I believe that it's preferable to have the sources(!) still be available in modules for logical reasons, but have a build-system gather/compile everything as part of the `install`/`update`/`clean` processes.

So at runtime, `zsh` would only read one precompiled file, but the drawback is of course reliance on the user properly running those commands above to keep their installations "fresh".

Eriner:
> the drawback is of course reliance on the user properly running those commands above to keep their installations "fresh"

This is easily resolved by inspecting file modification times on some lazily-loaded hook in the background or something. Could just default-on that automatic background cache-rebuilder. And if a user has turned on `noatime`, then "RTFM" is a sufficient response, assuming we put this into the documentation.

ericbn:
Oh, replacing all the `source foo.zsh` calls inside a separate `init.zsh` that is sourced by `.zimrc` by copying everything directly inside the `.zimrc` is a big one.

I still like the challenge of only relying only on Zsh and standard *nix tools to implement Zim. Also, we don't need our code to be fast if its purpose is to build a final all-in-one `.zimrc` and call git commands for installing/updating module repos.

If the build process is triggered after each install/update command, we'll cover the evey-day usage. User only need to manually call the build if they do other changes to their modules configuration in `.zimrc` (e.g. disable a module)

My only concern with having a all-in-one `.zimrc` is that we'll be touching a file that the user also edits manually. We can delimit the area that the build process touches with guard comment lines, I think that's fine. But I'm worried about making the file too big to be edited manually (user getting lost in so many lines added by the build and having a hard time finding his own custom code in the beggining and end of the file).

Eriner:
@ericbn vim folds are the only thing I can think of to solve that problem.

PatTheMav:
@ericbn just so I understand you right: The goal would be for the user to have a `.zshrc`, preferably source'ing one (assembled) `.zimrc` (and maybe additional files, e.g. Powerlevel10ks `.p10k.zshrc`,  but that's module dependent stuff), hence only one precompiled file to load as well?

Wouldn't moving all zim-related configuration into `.zshrc` be sufficient if the `zstyle`s and environment variables are set before sourcing it? The installation code might be tricky, but appending config settings + import to the bottom with an additional backup copy should be a good-faith approach.

ericbn:
@PatTheMav, the idea I'm working on is that `.zimrc` will not be sourced by `.zimrc` anymore. It should only contain module configurations, and will just be sourced by the `zimrc` tool before doing install/update of module repos, or re-building the initialization script.

@Eriner, just found an issue when trying to have an all-in-one initialization script: it breaks the `%N` escape, as used in at least [zsh-syntax-highlighting/zsh-syntax-highlighting.zsh](https://github.com/zsh-users/zsh-syntax-highlighting/blob/be3882aeb054d01f6667facc31522e82f00b5e94/zsh-syntax-highlighting.zsh#L39), since code is not running inside the original script file anymore. And setting the value of `0` does not work for `%N`.

Eriner:
@ericbn Yeah, this is exactly why I was considering trying to do this with a real programming language. This is a really tough thing to try to solve. We need to source files that are sourced by the file's we're sourcing, but after variable expansion, loops, conditionals, etc. 

```
# Resolve highlighters directory location.
_zsh_highlight_load_highlighters "${ZSH_HIGHLIGHT_HIGHLIGHTERS_DIR:-${${0:A}:h}/highlighters}" || {
```

I have a feeling there are going to be quite a few cases like this. I can't think of a better solution than monkey-patching the loaders for multi-plugin-file projects like that. Ideally, we do it by defining our own variable/function overrides and not by completely re-implementing the loader.

Perhaps the best way to do this is to have a .zimpatch file or something that we can add to the project repos on disk which define patches. This would let project owners also manage their own patches, which may save us time and hassle in the future.

PatTheMav:
FWIW I do believe when it comes to maintaining zim as a publicly available framework you have to be aware of diminishing returns when reducing usability/extensibility for speed.

The `develop` branch is kinda sexy, as it allows to add and run modules by just adding the GitHub path and run `zimfw install` and be done with it. It just works and as such is on par with Prezto, Antigen, Zplugin, Antibody, the lot (and one might even argue that it's even simpler since @ericbn removed dedicated `prompt` functionality).

That is - by itself - a big win. `zim` also seems to be the fastest of the lot even with all default modules enabled. And a fully-supported module only needs an `init.zsh` and can place custom functions in a subfolder of the same name. Easy.

My former executive producer always liked to say "perfect is the enemy of good", so maybe (just maybe) finalise the current state of the `develop` branch as a major version bump of zim (as it's not backwards compatible), face the deluge of GitHub comments coming our way (maybe prepare an FAQ based on feedback from the early adopters) and then brainstorm how to achieve the "holy grail" - maybe @romkatv has some ideas as he managed to do some "JIT"-like caching for https://github.com/romkatv/powerlevel10k that requires at least zsh 5.4.

At the very least it will resolve the "issue" of a dangling branch in limbo that has some very nice things going for it.. 😉

ericbn:
The new zimfw tool implementation is ready, with a more flexible zmodules configuration syntax (not using zstyle anymore FWIW), and generating a static init.zsh script. I'm finishing with the updates on the documentation, install script and templates, and on some of the modules, and I'll be ready to push the changes to the develop branch. These will be the last breaking changes to the develop branch, and then I'll be merging this to master as a new release. If you're using `zimfw upgrade` be aware that the next changes will break your zimfw tool, so another fresh install is recommended instead.

rjcoelho:
@ericbn can we get a .zimrc sample ? Thanks

ericbn:
@rjcoelho, here's a .zimrc sample:

```
zmodule environment
zmodule git
zmodule input
zmodule utility

zmodule sindresorhus/pure --source async.zsh --source pure.zsh

zmodule zsh-users/zsh-completions
zmodule completion
zmodule zsh-users/zsh-autosuggestions
zmodule zsh-users/zsh-syntax-highlighting
zmodule zsh-users/zsh-history-substring-search
```

And here's the zmodule usage:
```
Usage: zmodule <url> [-n|--name <module_name>] [options]

Repository options:
  -b|--branch <branch_name>  Use specified branch when installing or updating the module
  -t|--tag <tag_name>        Use specified tag when installing or updating the module
  -z|--frozen                Don't install or update the module

Startup options:
  -f|--fpath <path>              Add specified path to fpath
  -a|--autoload <function_name>  Autoload specified function
  -s|--source <file_path>        Source specified file
  -d|--disabled                  Don't use or clean the module
```
(the fpath, autoload, source and disabled options are new compared to the current develop code, made possible thanks to the other changes that come with them)

rjcoelho:
@ericbn it looks good/clean! Feel free to push to develop, I will happily test it🙂

Do we have dependencies between modules ? Does git depend on environment for example ? 

Also does it make sense to include zsh-defer to improve startup time ?

rjcoelho:
@ericbn why not simply git submodule add each plugin and zcompile them:

```
# If there is no *.zsh.zwc or it's older than *.zsh, compile *.zsh into *.zsh.zwc.
if [[ ! ~/zsh-autosuggestions/zsh-autosuggestions.zsh.zwc -nt ~/zsh-autosuggestions/zsh-autosuggestions.zsh ]]; then
  zcompile ~/zsh-autosuggestions/zsh-autosuggestions.zsh
fi
source ~/zsh-autosuggestions/zsh-autosuggestions.zsh
```

You could even defer it
`zsh-defer source-compiled ~/zsh-autosuggestions/zsh-autosuggestions.zsh`

romkatv:
> You could even defer it
> `zsh-defer source-compiled ~/zsh-autosuggestions/zsh-autosuggestions.zsh`

Or use powerlevel10k with instant prompt. It has several advantages over zsh-defer:

1. Initialization is done normally and not from zle. Initializing stuff from zle has many [problems](https://github.com/romkatv/zsh-defer/blob/master/README.md#caveats).
2. There are no keyboard input freezes during initialization. When using `zsh-defer`, zsh effectively freezes for the duration of each deferred action.
3. User input has deterministic effects. If you press a key, it'll always invoke the same zle widget regardless of when you press it. This is not the case when using zsh-defer.
4. You don't need to make any changes to zshrc other than adding the instant prompt preamble at the top. Using zsh-defer is more work.

You can still use zsh-defer even if you enable instant prompt but there are virtually no advantages (zsh already starts instantly) and quite a few disadvantages (see above).

rjcoelho:
@romkatv how do I defer a zsh-autosuggestions during powerlevel10k ?

romkatv:
> @romkatv how do I defer a zsh-autosuggestions using powerlevel10k ?

With `zsh-defer`. It works the same way with any theme, including powerlevel10k.

> It's a prompt afaik. I know it also defers but I don't even use that prompt.

It doesn't defer but it has a feature called [instant prompt](https://github.com/romkatv/powerlevel10k/blob/master/README.md#what-is-instant-prompt) which is better than deferring. See my previous comment for why it's better. Once you enable instant prompt, it becomes pretty much irrelevant how you load your plugins. There is no practical difference between the slowest plugin manager and the fastest with everything deferred. If someone were to add `sleep 1` to your `~/.zshrc` as a prank, it would take you a very long time to notice. You can still use `zsh-defer` but there is just no point in doing so.

Which theme/prompt do you use?

rjcoelho:
@romkatv I alternate between pure and lean

romkatv:
> @romkatv I alternate between pure and lean

Do you mean *Pure* and *Lean* styles in powerlevel10k? I understood your [previous comment](https://github.com/zimfw/zimfw/pull/323#issuecomment-560327949) as implying you **don't** use powerlevel10k. If you actually do, try enabling instant prompt and see how fast zsh starts. You can also try disabling deferred loading everywhere to verify that it really makes no difference.

> You can enable instant prompt either by running `p10k configure` or by manually adding the following code snippet at the top of `~/.zshrc`:
> 
> ```zsh
> # Enable Powerlevel10k instant prompt. Should stay close to the top of ~/.zshrc.
> # Initialization code that may require console input (password prompts, [y/n]
> # confirmations, etc.) must go above this block, everything else may go below.
> if [[ -r "${XDG_CACHE_HOME:-$HOME/.cache}/p10k-instant-prompt-${(%):-%n}.zsh" ]]; then
>   source "${XDG_CACHE_HOME:-$HOME/.cache}/p10k-instant-prompt-${(%):-%n}.zsh"
> fi
> ```

If by "pure" you meant https://github.com/sindresorhus/pure, you can switch to powerlevel10k and choose *Pure* style in the configuration wizard. Your prompt will look identical to pure but it'll be faster and with extra optional features such as instant prompt.

rjcoelho:
@romkatv i don't use powerlevel10k, and don't intend to switch to it. 

Now back to this thread. Why use a plugin manager (which is what zim fw's develop branch is ) when we can git submodule add + zcompile + defer it ?

romkatv:
> @romkatv i don't use powerlevel10k, and don't intend to switch to it.

Fair enough. I've only mentioned powerlevel10k because you've quoted a piece of my work (or two). I felt obliged to clarify that I myself don't use it because there is a better option by far.

> Now back to this thread. Why use a plugin manager (which is what zim fw's develop branch is ) when we can git submodule add + zcompile + defer it ?

I understand that this question is not directed to me but wanted to note that zimfw doesn't defer anything. Your question is thus threefold:

1. *Why use a framework such as zimfw rather than adding external git repositories as submodules to your own dotfiles repo?* The answer is that you definitely can and should add dependencies as submodules to your dotifles repo. One of those subodules can be zimfw, or prezto, or oh-my-zsh. Or all three. You don't have to choose between `git submodule` and zimfw. You can also make a fork of zimfw or prezto your top-level dotfiles repo. Prezto documentation even says that it's "highly recommended" that you fork it.
2. *What does zimfw bring in terms of zsh startup speedup other than calling `zrecompile`?* I think the answer is "nothing" but I may be wrong. I believe you can make prezto as fast as zimfw by adding a handful of lines to `~/.zpreztorc`. Naturally, using `zsh-defer` or instant prompt will blow these optimization out of the water.
3. *Why doesn't zimfw use deferred loading?* There is little reason in having builtin deferred loading given that you can use `zsh-defer` where you want it (e.g., `zsh-defer zmodule zsh-users/zsh-autosuggestions`). Plus, deferred loading has lots of issue and there is a much better solution to the problem of slow zsh startup.

ericbn:
@rjcoelho, the way I see it, which is different than oh-my-zsh and prezto do, is that adding plugins as submodules is a bad idea. This is a discussion that started [more than 3 years ago here](https://github.com/zimfw/zimfw/issues/88).

It's harder to manage an open source project with many submodules, and I mean it from the point of view of the maintainer of the project, not from the point of view of having a repo that I'll add to or fork as my personal dotfiles repo.

Look at some open issues we have that would be avoided by changing from an "open source project with plugins as submodules" to a "open source project that manages your plugins as separate repos, provided we maintain our own plugin repos in our organization, and welcome contributors who want to maintain their plugins in separate repos in our organization too": #345, #330, #324, #309, #269, #138, #33

Users prefer you tell them (using pure as an example, other examples can be even more complex):

> Add `zmodule sindresorhus/pure -s async.zsh -s pure.zsh` to your .zimrc and run `zimfw install`.

than

> Run:
> ```
> cd ${ZIM_HOME}
> git submodule add https://github.com/sindresorhus/pure.git modules/pure
> git commit # if you don't want to keep your repo dirty
> ```
> Add to your .zshrc:
> ```
> source ${ZIM_HOME}/modules/pure/async.zsh
> source ${ZIM_HOME}/modules/pure/pure.zsh
> ```

PatTheMav:
@ericbn Wil module configurations (e.g. autoloading certain ssh identities, double-dot-expansion, et. al.) also be handled by `zmodule`? The api doc above doesn't suggest it has switches for these things.

ericbn:
@PatTheMav, module configurations are still done with zstyle (or environment variables, depending on the module). The change is that this configuration moves from .zimrc to .zshrc (as .zimrc won't be sourced during startup anymore).

PatTheMav:
@ericbn Got it - so `zimfw` in essence becomes a module manager with on-demand compilation and some batteries (self-supplied modules) and an easy to maintain module framework included. Sounds good.

rjcoelho:
@ericbn is the branch/pr updated ? I can test it if you need it.

ericbn:
Zim is updated with the new breaking changes. Please reinstall it.

* .zimrc now contains only module definitions, that uses a new zmodule function, instead of zstyle.
* zimfw now generates a static init.zsh script, that is sourced by .zshrc
* .zimrc is not sourced during the shell initialization anymore, only the static init.zsh
* configuration that needs to be executed during shell initialization was moved to .zshrc
* directory and history modules are deprecated, and their code was moved to the environment module

rjcoelho:
@ericbn can we get install instructions on README alternative to running install script ? This is because I keep my dotfiles under git including zimfw as a submodule. And also because some people aren't comfortable sourcing shell scripts from the net. Thanks!!

jkaan:
> @ericbn can we get install instructions on README alternative to running install script ? This is because I keep my dotfiles under git including zimfw as a submodule. And also because some people aren't comfortable sourcing shell scripts from the net. Thanks!!

That would definitely help me as well in testing these changes!

README.md
```diff
@@ -3,96 +3,109 @@ Zsh IMproved FrameWork
...
-5. Open a new terminal and finish optimization (this is only needed once, hereafter it will happen upon desktop/tty login):
+Add `zmodule` calls to your `.zimrc` file to define the modules to be loaded.
+The modules are loaded in the same order they are defined. Add:
 
-       source ${ZDOTDIR:-${HOME}}/.zlogin
+    zmodule <url> [-n|--name <module_name>] [options]
 
-6. You're done! Enjoy your Zsh IMproved! Take some time to read about the [available modules][modules] and tweak your `.zshrc` file.
+where &lt;url&gt; is the required repository URL or path. The following formats
+are equivalent: *name*, zimfw/*name*, https://<em></em>github.com/zimfw/<em>name</em>.git
```
kaznovac:
```suggestion
are equivalent: *name*, zimfw/*name*, https://github.com/zimfw/<em>name</em>.git
```

ericbn:
@rjcoelho, @jkaan, I've added the manual installation steps to [README.md](https://github.com/zimfw/zimfw/blob/develop/README.md). But it still involves downloading a file and then sourcing it!  :- )

Suggestions on how to improve this are more than welcome. Also, as a security measure, we're enforcing that only signed commits are allowed for these scripts that are downloaded and sourced ([install.zsh](https://github.com/zimfw/install/commits/develop/install.zsh) and [zimfw.zsh](https://github.com/zimfw/zimfw/commits/develop/zimfw.zsh)).

ericbn:
@Eriner, I'm considering this done. There's still room for improvements, it's a good base for future enhancements, and more importantly the main decisions are here (after more than 3 years of thoughts).

Created #347 to discuss "Create an all-in-one initialization script?", as the current code can be merged independent from that, and that improvement can be done later if wanted.

rjcoelho:
@ericn in manual instructions step 3, intead of copy I prefer to git clone since then we can git pull for updates. Don't you agree ?

ericbn:
Added new module(s) to .zimrc? `zimfw install`
Removed module(s) from .zimrc? `zimfw clean-modules` 
Just changed startup options in module(s) in .zimrc? `zimfw build` 
Want to update your modules to their latest revisions? `zimfw update`
Want to upgrade zimfw.zsh to its latest version? `zimfw upgrade`
Want info about Zim and the system? `zimfw info` 

@rjcoelho, it would be nice to have suggestions on how to improve the documentation.

rjcoelho:
@ericbn add your answer to my last question (on what to call when we add/remove/update modules) to the README. That would improve docs.

ericbn:
@thiagokokada, you might find our new plug-in mechanism in the develop branch a useful replacement for your [zit](https://github.com/thiagokokada/zit) minimal plugin manager. Do we miss anything that zit supports?

(Also, we're deprecating some repositories you use, like directories and history)

thiagokokada:
@ericbn I will wait until the changes are merged on `master` to decide what to do.

I particularly quite like my current setup, and seeing the `develop` branch it seems that it is much more difficult to just import some `zim` files and source them. However if adopting zim is the only way forward I may change my setup, since I quite like the zim speed.

PatTheMav:
@ericbn Branch works fine so far - I will still have to adapt my Makefile's zsh installation target with the new manual installation steps (proper `.zshrc`, `.zlogin` and `.zimrc` are part of my dot files anyway), but otherwise it works as intended.

As for the maintenance commands, I do believe that `install`, `update` and `upgrade` follow the established standards set by apt/brew, but maybe aliasing `clean-modules` into `uninstall` might be worth a shot (with the necessary automatic cleanup/recompilation necessary).

ericbn:
@PatTheMav, that's great feedback! I'll completely rename `zimfw clean-modules` to `zimfw uninstall`, and have `zimfw clean` only do `clean-compiled` and `clean-dumpfile`. Semantically, it makes much more sense because we will be then cleaning temporary files than are later compiled/generated again, which is not the case for a module (which we'll be uninstalling now instead of cleaning).

As for using the latest Zim in your dotfiles, let me know if you see any issue. I'd suggest you add the ~/.zim/modules subdirectory to your .gitignore (maybe also the ~/.zim/init.zsh file), and let them be generated by `zimfw install` the first time you checkout the dotfiles repo to a new machine. That's only a suggestion from the point of view of the motivations that led to having zimfw.zsh as a single file, and maybe there are better ways that you prefer.

PatTheMav:
@ericbn - you're welcome. This way it should be more or less obvious how to manage modules as it follows other package managers' idioms.

As for my dotfiles, I don't manage zim inside of them, because once installed it takes care of itself. My Makefile has a `zsh` target that installs `zsh` as well as `zim`, so I just had to add the download and source parts of the manual install to it. The config files are symlinked automatically via the `install` target which is run as a precondition anyway.

ericbn:
Should we get rid of the UTF-8 characters to be friendlier with wide variety of environments, as suggested by @kaznovac [before](https://github.com/zimfw/zimfw/pull/323#issuecomment-455920179).

I think **yes**! The checkmark character is problematic even in terminals that try to support UTF-8, like the Windows terminal, and I've seen UTF-8 totally fail in FreeBSD for example.

I'm planing on replacing: (ellipsis) `…` by `...`, (failure) `✗` by `x`, (success) `✓` by `)`.

Screen Shot 2019-12-10 at 18 28 30

@Eriner, thoughts? Waiting for your approval of the whole change too so we can release this.

rafifos:
For some reason, GitKraken (and probably git hooks too) can't use zimfw. Shouldn't zimfw be already available since `.zlogin` is sourced after `.zshrc`?

I'm not sure if I'm missing something, I'm kinda of a beginner in shell (and ZSH has it's quirks too, which doesn't help much to understand it)

> GitKraken hook output (post-merge)
```sh
/Users/rafifos/.zlogin:4: command not found: zimfw
Hook process exited.
```

> Relevant files
```sh
# ~/.zshenv: Sourced on ALL invocations of ZSH(1), unless the -f option is
# set.
#
# NOTE: It should NOT normally contain commands to set the command search path,
# or other common environment variables unless you really know what you're
# doing. E.g. running "PATH=/custom/path gdb program" sources this file (when
# gdb runs the program via $SHELL), so you want to be sure not to override a
# custom environment in such cases.  Note also that .zshenv should not contain
# commands that produce output or assume the shell is attached to a tty.
#
export ZSH_HOME="${ZDOTDIR:-${HOME}}"
export ZIM_HOME="$ZSH_HOME/.zim"
```

```sh
# ~/.zshrc: Sourced by ZSH(1) in interactive shells. It should contain commands
# to set up aliases, functions, options, key bindings, etc.
#

# Installs and sources ZIM if it isn't installed.
if [[ ! -f ${ZIM_HOME}/zimfw.zsh ]]; then
  curl --create-dirs -fLosS "${ZIM_HOME}/zimfw.zsh" "https://raw.githubusercontent.com/zimfw/zimfw/develop/zimfw.zsh"
  source ${ZIM_HOME}/zimfw.zsh install
fi

# Update static initialization script if it's outdated, before sourcing it
if [[ ${ZIM_HOME}/init.zsh -ot ${ZSH_HOME}/.zimrc ]]; then
  source ${ZIM_HOME}/zimfw.zsh init -q
fi
source ${ZIM_HOME}/init.zsh
```

```sh
# ~/.zlogin: Sourced by ZSH(1) when starting as a login shell (--login).
#
# Initialize Zim
zimfw login-init -q &!
```

PatTheMav:
@ericbn I would vote for making it a switch/`zstyle` - many modern prompts rely on UTF-8 shenanigans a lot (e.g. Powerline glyphs, Nerd-Font glyphs), so UTF-8 support should be widespread.

The remaining question would be whether this `zstyle` defaults to `on` (which would make sense if there's a big cross section of `zimfw` users and those of modern prompts) or `off`.

PatTheMav:
@rafifos One way for that error to happen is to use a login (but non-interactive shell), e.g.  `zsh -lc 'echo "Test"'`. Hooks should usually run with the shell defined the shebang of the hook-script, so I'm surprised GitKraken invokes `zsh` at all..

Can you refer us to the post-merge hook script maybe?

ericbn:
> One way for that error to happen is to use a login (but non-interactive shell), e.g. `zsh -lc 'echo "Test"'`.

@rafifos, thanks for reporting the issue above. @PatTheMav, completely right, a non-interactive login shell will currently yield `command not found: zimfw`.

Maybe to best cover this, as its possible that some configurations will only open either non-interactive login shells (`l`) or interactive non-login shells (`i`), we should have a static login-init.zsh file that is generated too, so .zlogin could source it directly without depending on the zimfw function (as we have in the master branch). Let me work on this.

> I would vote for making it a switch/`zstyle` - many modern prompts rely on UTF-8 shenanigans a lot (e.g. Powerline glyphs, Nerd-Font glyphs), so UTF-8 support should be widespread.

@PatTheMav, I want to avoid having a configuration for this. I can already foresee an issue being opened because the user is not getting the UTF-8 characters right on their terminal. And I consider it a legitimate concern for users wanting to use Zim in "less modern" environments. The ASCII set gives us 94 visible characters to choose from, and everybody likes them.  :- )

PatTheMav:
@ericbn Yeah I would just make it ascii and call it a day. If you consider `zimfw` the bedrock of an extended `zsh` environment, it needs to make as few assumptions about features as possible. It already requires a quite modern `zsh` version, so all "fancy" requirements are better served by (user-provided) additional modules.

Maybe I'll just enhance/extend the tool in my own module to get back UTF-8 symbols, depends on how it's composed I guess.. ;)

rafifos:
> @rafifos One way for that error to happen is to use a login (but non-interactive shell), e.g. `zsh -lc 'echo "Test"'`. Hooks should usually run with the shell defined the shebang of the hook-script, so I'm surprised GitKraken invokes `zsh` at all..
> 
> Can you refer us to the post-merge hook script maybe?

@PatTheMav it's a [Husky](https://github.com/typicode/husky)-generated hook. But it probably is a GitKraken issue (to run ZSH instead of `sh`).

```sh
#!/bin/sh
# husky

# Hook created by Husky
#   Version: 3.0.5
#   At: 2019-12-9 15:42:26
#   See: https://github.com/typicode/husky#readme

# From
#   Directory: /Users/rafifos/<hidden>/node_modules/husky
#   Homepage: https://github.com/typicode/husky#readme
```

ericbn:
@Eriner, we have this [new issue above](https://github.com/zimfw/zimfw/pull/323#issuecomment-565189870) reported by @rafifos that I consider worth fixing before merging this branch. 

I'm thinking of the following solution, which involves changing the install dotfiles templates (so that's why I want to fix this before merging).

Add new zshenv template that contains:
```
ZIM_HOME=${ZDOTDIR:-${HOME}}/.zim
```
Change zlogin template to:
```
source ${ZIM_HOME}/login_init.zsh
```
so this way `ZIM_HOME` is defined even for non-interactive login shells (as zshrc is only sourced in interactive shells). Reference: http://zsh.sourceforge.net/Intro/intro_3.html

Then have zimfw generate the `${ZIM_HOME}/login_init.zsh` file during `zimfw build`, and change `zimfw compile` to source that file (instead of being a hardcoded function).

You'll need to check the current solution to see where I'm coming from for these changes, as in general terms they'll revert changes done in relation to the master branch. (I.e., we currently don't have a `login_init.zsh` file anymore, and the zlogin template is calling `zimfw login-init -q &!` instead, motivated by having just one core `zimfw.zsh` file, that defines the `zimfw` function).

As a note: this issue also affects the master branch, but it fails silently there because zlogin tests for `[[ -s ${ZIM_HOME}/login_init.zsh ]]` and that fails in non-interactive login shells as `${ZIM_HOME}` results in empty. The fix there would be much simpler: only adding the zshenv template.

ericbn:
Updated zimfw to generate a login_init.zsh file. This has a breaking change, as there's no `zimfw login-init` action anymore, so that will fail in your .zlogin script. Update your .zlogin code to:

    source ${ZIM_HOME}/login_init.zsh -q &!

And for that to work even in non-interactive login shells, add a new .zshenv script with:

    ZIM_HOME=${ZDOTDIR:-${HOME}}/.zim

Respective templates in the install scripts were updated already.

@rafifos, the issue your reported should be fixed with these changes.

rafifos:
Yeah, it's working fine.

And I've got a new issue on my WSL machine, zsh complains about insecure directories, not sure what we can do to fix this tho.

```sh
zsh compinit: insecure directories, run compaudit for list.
Ignore insecure directories and continue [y] or abort compinit [n]? y%

rafifos in ~ at DESKTOP-CAOIPDL
➜ compaudit
There are insecure directories:
/home/rafifos/.zim/modules/archive/functions
/home/rafifos/.zim/modules/git/functions
/home/rafifos/.zim/modules/utility/functions
/home/rafifos/.zim/modules/zsh-completions/src
/home/rafifos/.zim/modules/archive
/home/rafifos/.zim/modules/git
/home/rafifos/.zim/modules/utility
/home/rafifos/.zim/modules/zsh-completions
```

ericbn:
@rafifos, maybe it's really a permission issue. Did you check https://stackoverflow.com/a/13785716 ?

What surprises me is the fact that you got the error message, [since we use compinit -C](https://github.com/zimfw/completion/blob/38b8a1dbea4f3a8639d335205fd239def9923898/init.zsh#L18). Using compinit -i also should ignore the permission issues.

According to the [doc](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Use-of-compinit):

> For security reasons `compinit` also checks if the completion system would use files not owned by root or by the current user, or files in directories that are world- or group-writable or that are not owned by root or by the current user. If such files or directories are found, `compinit` will ask if the completion system should really be used. To avoid these tests and make all files found be used without asking, use the option `-u`, and to make `compinit` silently ignore all insecure files and directories use the option `-i`. This security check is skipped entirely when the `-C` option is given. 

Are you using something else than our completion module?

rafifos:
> @rafifos, maybe it's really a permission issue. Did you check [stackoverflow.com/a/13785716](https://stackoverflow.com/a/13785716) ?
> 
> What surprises me is the fact that you got the error message, [since we use compinit -C](https://github.com/zimfw/completion/blob/38b8a1dbea4f3a8639d335205fd239def9923898/init.zsh#L18). Using compinit -i also should ignore the permission issues.
> 
> According to the [doc](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Use-of-compinit):
> 
> > For security reasons `compinit` also checks if the completion system would use files not owned by root or by the current user, or files in directories that are world- or group-writable or that are not owned by root or by the current user. If such files or directories are found, `compinit` will ask if the completion system should really be used. To avoid these tests and make all files found be used without asking, use the option `-u`, and to make `compinit` silently ignore all insecure files and directories use the option `-i`. This security check is skipped entirely when the `-C` option is given.
> 
> Are you using something else than our completion module?

@ericbn Nothing I can think of, since it's working fine on my A1398.

rjcoelho:
If I use `zmodule zdharma/fast-syntax-highlighting ` I get:
`re-compiling fast-syntax-highlighting.zwc: zrecompile:21: parse error near `)'
zrecompile:zcompile:133: can't read file: fast-syntax-highlighting/LICENSE`

I think we should only compile .zsh and .sh files

ericbn:
@rjcoelho, that error is because the plugin is adding its root dir (with a lot of files that are not Zsh function files) to the fpath, [here](https://github.com/zdharma/fast-syntax-highlighting/blob/c39c4675a679c0c1761b9bc908e49348fa76d837/fast-syntax-highlighting.plugin.zsh#L51).

One should be able to compile all files in the fpath directories, as we can see in the [zrecompile doc](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Recompiling-Functions):

>The following shell loop is an example of a method for creating function digests for all functions in your `fpath`, assuming that you have write permission to the directories:
>
>```
>for ((i=1; i <= $#fpath; ++i)); do
>  dir=$fpath[i]
>  zwc=${dir:t}.zwc
>  if [[ $dir == (.|..) || $dir == (.|..)/* ]]; then
>    continue
>  fi
>  files=($dir/*(N-.))
>  if [[ -w $dir:h && -n $files ]]; then
>    files=(${${(M)files%/*/*}#/})
>    if ( cd $dir:h &&
>         zrecompile -p -U -z $zwc $files ); then
>      fpath[i]=$fpath[i].zwc
>    fi
>  fi
>done
>```

so we can take advantage of compiled function digest files, as explained in the [autoload doc](http://zsh.sourceforge.net/Doc/Release/Functions.html#Autoloading-Functions):

>For each *element* in `fpath`, the shell looks for three possible files, the newest of which is used to load the definition for the function:
>
>`element.zwc`
>
>A file created with the `zcompile` builtin command, which is expected to contain the definitions for all functions in the directory named *element*. The file is treated in the same manner as a directory containing files for functions and is searched for the definition of the function. If the definition is not found, the search for a definition proceeds with the other two possibilities described below.
>
>...

You can get the same error you reported by doing:
```
% autoload -Uz LICENSE
% LICENSE
LICENSE:21: parse error near `)'
```

As a note: "Prezto"-type plugin are much more flexible in this sense than "Oh-My-Zsh"-style plugins:
* run `init.zsh`, if it exists (no mess with i-put-whatever-name-i-want-here.plugin.zsh)
* if `functions` subdirectory exists, add it to fpath, and call autoload for all files in that directory.

You might want to report the issue upstream.

rjcoelho:
@ericbn we need to review README. For example the first line "zimfw is a configuration framework" .. it's more like a plug-in manager. Also what's let to close a release to master ? We need a stable release so more users can report bugs/features.

MeisterP:
completion for `zimfw` command would be nice. maybe as part of https://github.com/zimfw/completion?

ericbn:
Added warning in 7e7e57c for current master branch users about Zim 1.0.0, before this gets merged.

ericbn:
Will merge this on Jan 7 2020 to give master branch users enough time to get the commit with the warning message.

jkaan:
I was trying to switch to this new version today and I noticed that after installing [pure](https://github.com/sindresorhus/pure) I got some errors relating to async, so I installed those: https://github.com/mafredri/zsh-async. But after that I still had errors telling me that it couldn't find some functions `async_*`.

I can give some more detailed logs if needed, did someone get the pure theme to work? Or does someone have an equivalent which looks exactly like Pure 😊 . I vaguely remember as well that I read in this thread that that particular theme wouldn't work anymore.

ericbn:
@jkaan, you picked the two very well known plugins that are not automatically detected by frameworks, and that's why they have [complex](https://github.com/sindresorhus/pure#integration) [installation](https://github.com/mafredri/zsh-async#integration) instructions.

It's unfortunate that there's no official standard for plugins, and that it's the Wild West out there. And that's why `zmodule` has [documented parameters to override the automatic detection](https://github.com/zimfw/zimfw/blob/develop/README.md#zmodule) for these edge cases. After you read that documentation, and after you take a look at the [sindresorhus/pure](https://github.com/sindresorhus/pure) and [mafredri/zsh-async](https://github.com/mafredri/zsh-async) repos, I hope it gets clear that you'll need to do:

* [sindresorhus/pure](https://github.com/sindresorhus/pure):

   ```
   zmodule sindresorhus/pure --source async.zsh --source pure.zsh
   ```
   (no need to include mafredri/zsh-async then)
* [mafredri/zsh-async](https://github.com/mafredri/zsh-async):

   ```
   zmodule mafredri/zsh-async --name async
   ```
   or
   ```
   zmodule mafredri/zsh-async --source async.zsh
   ```
   (if you have any of the two lines above, then you can just have `zmodule sindresorhus/pure` for pure)

I'll make sure to add a PR to sindresorhus/pure to update the Zim installation instructions on Jan 7, when this branch will get merged to master.

jkaan:
@ericbn Thanks for the fast reply, after doing this everything works smoothly!

AtomicCoding:
I've switched to the develop branch about a week ago, and everything so far has been amazing, it's really great to see how far Zim has come. However, I've been having a mild issue with how Zim precompiles things into one script. I have this line in my .zimrc:
```if (( ${+TERM_PROGRAM} )); then
if [ $TERM_PROGRAM = "iTerm.app" ]; then
zmodule romkatv/powerlevel10k
else
zmodule minimal
fi
else
zmodule minimal
fi```
And the expected behavior would be that when I run iTerm (which has a nerd font installed), I'd get a powerlevel10k prompt but when SSHing or using any other terminal, I'd get minimal. However, due to the way it seems to be compiled (I'm assuming it just checks which modules have been defined, and loads them), I've had to run `zimfw update` each time I switch from different terminals to make sure the prompt changes. This isn't a significant issue, and would likely require some work to copy over to the pre-compiled script, but it's just something I've been experiencing.

PatTheMav:
@AtomicCoding Yeah if I'm not mistaken that kind of dynamic loading isn't really possible with new `zim` anymore as the module configuration is transpiled into an `init.zsh` containing all necessary `fpath`s, `autoloaders and `source` calls with the corresponding modules fetched from Github and pre-compiled.

As far as I can see there is no way to "select" both modules, but _not_ automatically sourcing them (because that's what `zim` will do, as the prompts will have `.zsh-theme` files) so you could have both managed by `zim`, but then `source` them yourself depending on `$TERM` e.g. in your `.zshrc`.

@ericbn any idea here? The most-straightforward option would be an option for `zmodule` that excludes them from autoloading (e.g. `--source /dev/null`). Bit of an edge-case though..

The only other way would be for a custom module that handles the switch based on `$TERM` and downloads/manages/sources the themes inside its own scope.

ericbn:
Done! 😌

thiagokokada:
Thanks @ericbn. I migrated my setup to use the recent changes, it was actually very easy: https://github.com/thiagokokada/dotfiles/commit/c7de37e69a42c55c9f180226b3616ad401c205c4.

Really like the split of `zimfw` (that seems to be mostly related to module management) and the rest (that are actual modules) :+1: .

I was thinking of migrating to use pure `zim` instead of `zit` however I still need to evaluate the changes. I kind like how minimal `zit` is, even if I do miss some features sometimes (that I don't implement to avoid feature creep). And it seems that `zim` now depends on Ruby and I don't know exactly how I feel about it.

Still, great work!

ericbn:
@thiagokokada, I'm only using Ruby to build the final zimfw.zsh file using ERB. But you definitely don't need Ruby to use Zim.

PR #328, opened 2019-01-28
Make ls working on osx

marek2901:
# Mac OX :( ls lacks some features

```
➜ .zim master ✓ ls --group-directories-first --color=auto
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -17,8 +17,10 @@ if (( terminfo[colors] >= 8 )); then
     else
       export LS_COLORS='di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43'
     fi
-
-    alias ls='ls --group-directories-first --color=auto'
+    # OSX ls doesn't understand group-directories-first and color=auto as valid params
+    if [[ ${OSTYPE} != darwin* ]]; then
+      alias ls='ls --group-directories-first --color=auto'
+    fi
   else
     # BSD
 


marek2901:
It seems like this one makes a fix, although there might be folks replacing osx ls with gls from homebrew so what I think there are 2 options really :).

1. Do the alias if the ls command (no matter alias or os type returns 0)
```
if ls --group-directories-first --color=auto > /dev/null 2>&1; then
  echo 'alias here'
fi
```
2. Or leave the condition for os check as is, and actually check for gls binary inside /usr/local/bin (homebrew bin) and do smt like this
```
if [[ current check ]]; then
  aliast "ls ...
elif [[ "is osx" && -x /usr/local/bin/gls ]]; then
  alias ls="gls --....
fi
```
the second option will automatically switch for gls in case there is one in the system

marek2901:
@ericbn Thanks for pointing out. I didn't notice it was homebrew change. So I'm closing this. Thanks !!

marek2901:
For now I made myself a work around
https://github.com/marek2901/config_files/blob/master/.bin/ls

BTW. `chmod` is also broken

anger

ericbn:
Oh! @marek2901, can you please change how `chmod` is also broken? I want to add that as another argument in the homebrew discussion.

ericbn:
Yeah, this is because we do:
```
if (( ${+commands[dircolors]} )); then

  alias lx='ll -X' # long format, sort by extension

  # Always wear a condom
  alias chmod='chmod --preserve-root -v'
  alias chown='chown --preserve-root -v'
fi
```
Another good catch! Looks like we'll have to make our script smarter so it can work in even this craziness that the homebrew contributors are enforcing.

marek2901:
Also, depending on a homebrew coreutils "revert" PR outcome, we can do smth like this:
https://github.com/marek2901/config_files/blob/master/.zshrc#L4

I'd rather wait for the revert (https://github.com/Homebrew/homebrew-core/pull/36494) to happen.

marek2901:
Better fix introduced in #329. Closing

PR #329, opened 2019-01-29
[utility] Don't test colors and smarter "if it's GNU"

ericbn:
Looks like most (if not all) others, will report at least `8`. See https://gist.github.com/michael-lazar/4e1beb7be279c72f59662127b5a234e6 (423 different terminals reporting 8, 15, 16, 52, 64, 88, or 256).

To test "if it's GNU", don't just test for the presence of `dircolors`, but also check if `ls` is GNU by accepting the `--version` parameter.

This is being done because MacOS's homebrew allowed installing a mixture of GNU and BSD tools together, where we got the GNU `dircolors`, but the BSD `ls`. See Homebrew/homebrew-core#36241

The homebrew changes were partially reverted in Homebrew/homebrew-core#36494, which does not mix the two tools above anymore, thus not breaking our script anymore either. But I prefer we keep this safer for the future.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -2,54 +2,57 @@
 # Utility Functions and Options
 #
 
-#
-# Colours
-#
-
-if (( terminfo[colors] >= 8 )); then
+if (( ${+commands[dircolors]} )) && ls --version &>/dev/null; then
+  # GNU
 
   # ls Colours
-  if (( ${+commands[dircolors]} )); then
-    # GNU
-
-    (( ! ${+LS_COLORS} )) && if [[ -s ${HOME}/.dir_colors ]]; then
-      eval "$(dircolors --sh ${HOME}/.dir_colors)"
-    else
-      export LS_COLORS='di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43'
-    fi
-
-    alias ls='ls --group-directories-first --color=auto'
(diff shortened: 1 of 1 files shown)


ericbn:
@marek2901, this is the change I propose to the utility module regarding homebrew's coreutils. Looks like homebrew is going to revert the piece of their change that broke our script, but I prefer to keep this safer for the future.

Closes #328

modules/utility/init.zsh
```diff
@@ -8,23 +8,27 @@
...
     else
       export LS_COLORS='di=1;34:ln=35:so=32:pi=33:ex=31:bd=1;36:cd=1;33:su=30;41:sg=30;46:tw=30;42:ow=30;43'
     fi
-
     alias ls='ls --group-directories-first --color=auto'
+    alias lx='ll -X' # long format, sort by extension
+
+    # Always wear a condom
+    alias chmod='chmod --preserve-root -v'
+    alias chown='chown --preserve-root -v'
```
ericbn: Hmm, this should actually not be here, because it's inside the `(( terminfo[colors] >= 8 ))` section...

ericbn:
@Eriner, what do you think?

PatTheMav:
FYI, to make this work "as intended" on macOS, users might need to symlink `gls` and `gdircolors` to non-prefixed names somewhere in their path if they don't choose to go full breakage with `coreutils`. Dunno if this _should_ be explained in the docs somewhere though.

ericbn:
Maybe we should test for NO_COLOR, or for the terminal type instead. See https://no-color.org

ericbn:
Closing this in favor of https://github.com/zimfw/utility/pull/1

PR #333, opened 2019-03-21
feat(git): add aliases for merge sign and verify

sQVe:
Adds the following aliases:
  - gmS: GPG-sign the resulting merge commit.
  - gmV: Verify that the tip commit of the side branch being merged is
  signed with a valid key, i.e. a key that has a valid uid: in the
  default trust model, this means the signing key has been signed by a
  trusted key. If the tip commit of the side branch is not signed with a
  valid key, the merge is aborted.

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -112,9 +112,12 @@ Aliases
 
   * `gm` joins two or more development histories together.
   * `gmC` performs the merge but does not commit.
-  * `gmF` performs the merge generating a commit even if the merge resolved as a
-    fast-forward.
+  * `gmF` performs the merge generating a commit even if the merge resolved as
+    a fast-forward.
   * `gma` aborts the conflict resolution, and reconstructs the pre-merge state.
+  * `gmS` commits with GPG signature.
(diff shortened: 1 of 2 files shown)


ericbn:
Thanks for the contribution, @sQVe!

Since `gmv` is not taken yet, I prefer we use that instead of `gmV` (since 926627652f4d3f122a1ae6645a11517c1043d86d, `v` is used to verify signatures, and `s`/`S` is used to sign). And can you please update the README accordingly too?

sQVe:
@ericbn :wave:

I was a bit quick about this yesterday, apologies for not being a bit more thorough. I have now added the proper information to the README and changed `gmV` to `gmv`. I'm leaving `gmS` unchanged, even though `gms` is free for use, so that the signing aliases are consistent.

ericbn:
I just pushed a couple of commits I had done before in [zimfw/git](https://github.com/zimfw/git). One of them adds the ability to configure the `g` prefix, so all aliases now have

    -alias g...
    +alias ${gprefix}...

I'll resolve those conflicts in your PR.

PR #334, opened 2019-03-22
feat(git): add alias for reflog

sQVe:
:wave: :point_up: 

Title pretty much says it all. PR adds alias for reflog.


ericbn:
I did a couple of commits to [zimfw/git](https://github.com/zimfw/git) and didn't copy them to zimfw/zimfw. One of them was adding this alias. I just pushed them.

PR #337, opened 2019-05-09
[template] expand note of highlighter conflicts for alacritty

kaseiwang:
Alacritty has #28 issue too.

### DIFF
--- a/templates/zimrc
+++ b/templates/zimrc
@@ -72,7 +72,7 @@ zinput_mode='emacs'
 # This determines what highlighters will be used with the syntax-highlighting module.
 # Documentation of the highlighters can be found here:
 #   https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md
-# For (u)rxvt, termite and gnome-terminal users,
+# For (u)rxvt, termite, alacritty and gnome-terminal users,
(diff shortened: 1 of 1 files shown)


ericbn:
I'm actually in favor of removing the `cursor` highlighter. Does it do any good in any situation? Not needed in MacOS iTerm or Terminal either.

@Eriner, thoughts?

kaseiwang:
Or just remove it from default config?

ericbn:
@Eriner, @Plloi, @PatTheMav, what do you think about removing the cursor highlighter from the zimrc template? ☝️

PatTheMav:
@ericbn Personally I'm fine with removing it as I use near exclusively use macOS, so I'm not affected. 

Also I've yet to see this specific functionality of `zsh-syntax-highlighting`, so you'd have my vote to remove it from default (!) settings - people can still add it if necessary.

Plloi:
I pulled it from my config due the issues it causes, so i may be a little biased... But I say pull it.

ericbn:
`cursor` highlighter removed.

@Plloi, any news from @Eriner?

Plloi:
...about what? though any contact i have is have is here in the comments.

ericbn:
Oh sorry, I don't know why I had the impression you two worked together or something like that :blush:

PR #226, opened 2017-11-21
[prompt] eriner: Add help, preview, parameters

ericbn:
and refactor code, inlining auxiliary functions inside `prompt_eriner_main`.

Theme colors are now customizable with parameters. Also, it works better with light backgrounds by using `%Sstandout%s` for the segments which were originally using black as foreground color (intended for dark backgrounds).

The help explains the parameters, and the preview shows one color variation.

PR #338, opened 2019-06-21
Added option to select foreground color for eriner theme

v1z4:
I had a problem with eriner theme. I use dark color mode, and it does not allow me to select black foreground for the status panel. I fixed it and added the option to select the foreground color. 

Usage:
`prompt eriner black blue green yellow black`

Now it looks like this:
Снимок экрана 2019-06-21 в 21 27 33

I think it's also possible to make an automatic switch to black foreground if user has been selected cyan/yellow/red and so on.

### DIFF
--- a/modules/prompt/themes/eriner.zsh-theme
+++ b/modules/prompt/themes/eriner.zsh-theme
@@ -20,10 +20,10 @@ prompt_eriner_help () {
   cat <<EOH
 This prompt is color-scheme-able. You can customize it using:
 
-    prompt eriner [status_color] [pwd_color] [git_clean_color] [git_dirty_color]
+    prompt eriner [status_color] [pwd_color] [git_clean_color] [git_dirty_color] [foreground_color]
 
 where the parameters are the background colors for each segment. The default
-values are black, cyan, green, and yellow.
+values are black, cyan, green, yellow, default
 
 In order for this prompt to render correctly, a font with Powerline symbols is
 required. A simple way to install a font with Powerline symbols is to follow
@@ -35,6 +35,7 @@ prompt_eriner_main() {
   local prompt_eriner_retval=${?}
   local prompt_eriner_color1=${1:-black}
   local prompt_eriner_color2=${2:-cyan}
+  local prompt_eriner_fg=${5:-default}
 
   ### Segment drawing
   # Utility functions to make it easy and re-usable to draw segmented prompts.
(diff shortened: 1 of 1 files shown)


ericbn:
We use standout for the foreground color, so the default should work in both light and dark terminal color schemes. This was done in #226.

Just double checked with Solarized Light and Solarized Dark:

Screen Shot 2019-06-21 at 15 18 39
Screen Shot 2019-06-21 at 15 14 59

Might be something wrong with your terminal color configuration.

vizakenjack:
Could you check cyan/yellow/green colors? 

For example, current zimfw and default iTerm2 Solarized Dark theme:
Снимок экрана 2019-06-23 в 15 31 13

Same iTerm2 theme and custom foreground color:
Снимок экрана 2019-06-23 в 15 35 44

ericbn:
We're assuming the first segment will have a dark background and the others a light background, on dark color schemes (or the other way around for light ones). Your settings allow configuring the foreground of the first segment, but not the others. It is still partial, and not sure if allowing setting the foreground of each segment separately won't be overkill.

Ideally the code should be able to decide or not to standout the foreground color based on the lightness of the background color, but I don't know if that's possible within the shell.

vizakenjack:
I think we can check for one of 8 standard colors: black, blue, green, yellow, cyan, white, magenta, red.

If it's black background, we need to set white/default foreground. Otherwise, we can use black foreground. 

I have tried to use this code to set fg color and it works fine:
  prompt_foreground_color() {
    if [[ $1 = "black" ]]; then
      prompt_eriner_fg="white"
    else
      prompt_eriner_fg="black"
    fi
  }

The only problem - I don't know how to compose all the segments. There is "%S" and I don't know how to replace it.

ericbn:
Here's a code that replaces both prompt_eriner_segment and prompt_eriner_standout_segment functions: 

```
  prompt_eriner_segment() {
    if [[ ${1} == black ]]; then
      print -n "%K{${1}}"
      [[ -n ${prompt_eriner_bg} ]] && print -n "%F{${prompt_eriner_bg}}%f"
      print -n "${2}"
    else
      print -n "%S%F{${1}}"
      [[ -n ${prompt_eriner_bg} ]] && print -n "%K{${prompt_eriner_bg}}%k"
      print -n "${2//\%F/%K}%s"
    fi
    prompt_eriner_bg=${1}
  }
```

I just don't like having to replace %F by %K when in standout, because the status segment uses %F in its code, as we assumed its background was always going to be dark.

vizakenjack:
Maybe it would be better to split this theme to eriner-dark and eriner-light ?

PR #341, opened 2019-09-25
Verify terminfo before bindkey

gfguthrie:
Check `terminfo` has a value before `bindkey` to prevent errors:

```
[84786]: /Users/xxx/.zim/modules/history-substring-search/init.zsh:bindkey:14: cannot bind to an empty key sequence
[84786]: /Users/xxx/.zim/modules/history-substring-search/init.zsh:bindkey:15: cannot bind to an empty key sequence
```


ericbn:
Good catch! Thanks @gfguthrie!

PR #348, opened 2019-12-08
Add init.zsh to gitignore

rjcoelho:
These are temp files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.zwc
 *.zwc.old
 custom/
+init.zsh
+init.zsh.old


ericbn:
The gitignore is not included in the installation, and the repo is not intended to be used in personal installations as a submodule.

This will only serve developers of zimfw.zsh. If that's the intention, I suggest we also include `modules/` here.

rjcoelho:
So why are *.zwc and *.zwc.old and included on gitignore ?

ericbn:
Okay, actually Zim modules have an init.zsh file, so we don't want to ignore them. E.g. ~/.zim/modules/git/init.zsh

Besides that, Zim builds a static initialization script, that initializes all enabled modules, in ~/.zim/init.zsh. You might want to commit this file to your dotfiles repo.

Also, the Zim core is just ~/.zim/zimfw.zsh, which can upgrade itself with the `zimfw upgrade` command. This is not new, [vim-plug](https://github.com/junegunn/vim-plug) and [antigen](https://github.com/zsh-users/antigen) work this way, to name two examples that come to my mind right now. So you just need:

```
~
├── .zim/  <-- this is NOT a Git repo
│   ├── init.zsh  <-- static script that initializes all your enabled modules
│   ├── zimfw.zsh  <-- the Zim core
│   └── modules/
│       ├── module1/  <-- this is a Git repo
│       ├── module2/  <-- this is another Git repo
│       ├── moduleN/  <-- and so on..
```

The advantage of this is that you don't have the situation of "Git repos inside a Git repo", as we had when ~/.zim was also a Git repo.

If you have a dotfiles repo, now the modules are directly your submodules, and that's it. **The Zim code is now just the zimfw.zsh, and you should not clone the whole zimfw repo, because all you need is the zimfw.zsh file, and because zimfw.zsh can be upgraded with `zimfw upgrade`. No need to clone the zimfw repo.** I don't know how to stress this enough. (Também posso dizer isso em português se quiser: Você não precisa de todo o repositório zimfw, só do arquivo zimfw.zsh, que pode ser atualizado usando o comando `zimfw upgrade`. O resto do repositório na verdade são só arquivos src usados para construir o zimfw.zsh, e o README.md e a licença MIT. Bem lembrado, talvez devessemos incluir a licença MIT como um comentário dentro o zimfw.zsh)  :- )

rjcoelho:
Understood. Then we need a way to say from which repo your zimfw upgrade pull. How can I fork it to improve it ?

ericbn:
I thought it was clear that zimfw.zsh comes from the develop branch of this (the zimfw/zimfw) repo. It's even in the manual installation steps:

> Copy https://raw.githubusercontent.com/zimfw/zimfw/develop/zimfw.zsh to ~/.zim/zimfw.zsh.

The remaining files in the branch are in fact just the src files used to build zimfw.zsh, and the README.md and the MIT license (which is now included as a comment in zimfw.zsh).

I nice way to contribute is maintaining and owning a new module in the zimfw organization. Maybe some code you already own in your dotfiles that would benefit other users too.

rjcoelho:
Thanks, I will close this PR

PR #357, opened 2020-01-07
Add CODE_OF_CONDUCT.md

ericbn:
We need to fill in the `[INSERT CONTACT METHOD]`.

@Eriner, should we put both our email addresses here, or do you suggest something else?

(Taken from the original at https://www.contributor-covenant.org/version/2/0/code_of_conduct.md)


Eriner:
@ericbn yeah, `m@tthamilton.com`

PR #358, opened 2020-01-08
add .zprofile to the list of startup files

kaznovac:
zim does not use/modify `.zprofile` in it's templates.
for completeness/performance the `.zprofile` should be compiled/cleaned if present.

ref: [zsh startup files](http://zsh.sourceforge.net/Intro/intro_3.html)

### DIFF
--- a/src/zimfw.zsh.erb
+++ b/src/zimfw.zsh.erb
@@ -5,7 +5,7 @@ class Zim
   def initialize
     @home = "${ZDOTDIR:-${HOME}}"
     @min_zsh_version = "5.2"
-    @startup_files_glob = ".z(shenv|shrc|login|logout)"
+    @startup_files_glob = ".z(shenv|profile|shrc|login|logout)"
     @version = "1.0.0"
     @ellipsis = " ..."
     @okay = "%F{green})%f "


ericbn:
Can you please also run `make` and include a short description of the change under Unreleased in CHANGELOG.md? 

I still need to create  CONTRIBUTING doc with this info.

ericbn:
Merged into develop with the extra changes.

kaznovac:
thank you @ericbn (sorry didn't find time to make requested changes)

p.s.
it's a bit more chore to commit to this new structure of zimfw src.
maybe introduce a ci bot for building master branch and tagging releases (in this case zimfw can be downloaded from releases and repository can remain clean without build artifacts)

ericbn:
Yeah, I agree there's room for automation here. It's just not totally clear to me what the process would look like yet. Also, I worry about automating when we don't have any unit tests.

The install script is already downloading the latest release asset instead of the raw file from the master branch. Maybe we should have the same in `zimfw upgrade`. Let's see how that goes.

ericbn:
I just tried gitpod.io/#https://github.com/zimfw/zimfw and it works like a charm!

Issue #400, opened 2020-06-06
Installing zim with .zimrc already there


PierreMarchand20:
Description
-----------

I am using zim with [yadm](https://yadm.io), that I would describe as git, but for your home repository. The issue I have is that I put my `.zimrc` and `.zshrc` files in my repo yadm/git, but then, when I try to install zim, it fails because it considers there is already a zim installation.

Is there a way to circumvent this ? Do I need to use the install script if I put zim as a submodule and I run `zimfw install` instead for example ?

By the way, I guess it is something you just added from #398, but I am not sure that failing is the right call. For example, yadm has a bootstrap function (see [here](https://yadm.io/docs/bootstrap#)), kind of a script that installs everything, in my CI I test it to be idempotent. But then it will automatically fail. For me the idea that it must be idempotent comes from the fact that if you just change one thing in you installation script, you can run it and it will only update the part you changed without breaking the rest (because it is idempotent). 

To quote yadm documentation: "It is best to make the logic of your bootstrap idempotent—allowing it to be re-run in the future when you merge changes made on other hosts." Note that it can apply to any installation script for dotfiles, it is not specific to yadm I think.

Steps to reproduce
------------------

1. Have a `.zimrc` or a `.zshrc` with some zim options before installing zim.

Images or other information
---------------------------

You can find my CI where the issue appears [here](https://travis-ci.com/github/PierreMarchand20/dotfiles/jobs/344424831)

ericbn:
Hi @PierreMarchand20. If you already have your .zimrc and .zshrc files, then it would also not make sense to have the install prepending the Zim configuration to these files, as I imagine you already have it there, right? In this sense, your issue is similar to #398.

Maybe a solution is to skip the step of prepending the Zim configuration when the install detects that Zim is already configured, instead of aborting. Does this make sense and solve your issue? Any different suggestion?

PierreMarchand20:
If the install process does something else than prepending the Zim configuration to these files, then yesI agree! But may be someone will want the current behaviour (be warned that there are still some Zim configuration to delete to get a fresh install). So may be a flag in the install script to not prepend these files ? 

So may be there could be some default behaviour (prepending or not) and a flag to do the opposite.

PierreMarchand20:
So I just struggled with a bug in my configuration (see #401) trying to overcome this issue in the meantime, and I understood that `.zshrc` will need to be compiled. Is this something that the install process will do, even if it does not prepend the the files ?

[...]

PR #402, opened 2020-06-12
Update manual install steps

ericbn:
to be simpler, and to teach users how to do it from outside Zsh. Users might use provioning scripts to manage their dotfiles, and knowing how to install from outside Zsh might be helpful in some cases. See #400.

@Eriner, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,26 +55,22 @@ If you have a different shell framework installed (like oh-my-zsh or prezto),
 
 ### Manual installation
 
-1. Start a Zsh shell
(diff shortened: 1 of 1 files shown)


ericbn:
@PierreMarchand20, does `zsh ~/.zim/zimfw.zsh install` also works for you?

PierreMarchand20:
> @PierreMarchand20, does `zsh ~/.zim/zimfw.zsh install` also works for you?

Yes, it works!

PR #440, opened 2021-10-26
Update README.md

Konfekt:
Similar to https://github-wiki-see.page/m/junegunn/vim-plug/wiki/tips

Useful if the .zim folder is not version tracked in a dotfiles repo.


ericbn:
This will work with the new version 1.6.0 of zimfw, that was just released. Thanks for the contribution!

PR #450, opened 2022-01-07
Deprecate compile in background via login_init.zsh

ericbn:
Changes are:
  * Don't compile in the background anymore, only via the `zimfw` tool after actions where scripts can change (install, update, etc.)
  * Don't compile users startup scripts anymore (.zshenv, .zshrc, etc.)
  * Make output of `zimfw init` friendlier for the terminal startup screen when called without `-q`.

See https://github.com/romkatv/zsh-bench/pull/11#issuecomment-994979683


ericbn:
@Eriner, since this changes one of the original Zim designs, I'd like to have your approval on this one.

ericbn:
These are numbers from the [GitHub Action run](https://github.com/zimfw/zsh-framework-benchmark/actions/runs/1670132831):

framework|mean|stddev|min|max
---|---|---|---|---
zim (latest release)|0.060|0.002|0.059|0.070
zim-branch (this PR)|0.061|0.010|0.059|0.155

results

Maybe we can consider the numbers equivalent, and hopefully the change makes Zim even more user-friendly.

Eriner:
I have no problem with this. The compilation was originally placed in zlogin with linux users in mind. Mac OS considers every shell a login shell, so this approach was flawed from the beginning.

My only issue is in the compilation of user files. My understanding is that Zsh by default reads the configuration file with the most recent mtime. If this is the case, I don't have a problem with moving the zcompile out of zlogin, but the functionality should remain unchanged.

If that isn't the case and Zsh reads the zwc regardless of mtime, then for correctness it should be left to the user to optimize. That said, my (admittedly poor) recollection is that Zsh inspects mtime and will read the source file if it is newer than the zwc.

The corruption on disconnect is what I would consider to be an edge-case, regardless of systemd prevalence. The change would make sense if we were still running the zcompile in zlogin, but if it is being moved to installation then "wait while this finishes" is implied and corruption is far less likely to occur.

Either way, LGTM, but those are my 2c.

Eriner (approved):

ericbn:
Cool. There are other reasons why compiling the user rc files is bad, also according to @romkatv, [here](https://github.com/romkatv/zsh-bench#cutting-corners).

I'm choosing to only compile the modules' scripts at least for the reason that the compile will not happen so ofter anymore -- it will happen only when the user calls `zimfw` to install/update the modules. So it makes more sense to only compile the files that `zimfw` has control over changes...

PR #465, opened 2022-06-28
Add Fig as an installation method to the README

brendanfalk:
We recently listed `zimfw` on [Fig Plugin Store](https://fig.io/plugins), so we'd love to have Fig listed as a download method on your readme. Thanks so much and please let me know if you have any questions!


sofyanox12 (approved):
Thanks! this could be very useful.

PR #516, opened 2023-08-09
Make the Usage section more friendly with a TL;DR

yuriaps:
added TL;DR


yuriaps:
I just saw that in the webpage of the módulos there are the instructions to install, bit still a bit confusing that in the Usage session there are only technical details about how it works and not easy way to use

